PR #9, opened 2016-01-26
Feature/update doc conventions

taroth21:
Team: Please have a look and provide feedback. I'd like to used this as a blueprint for updating the same file in our other doc repos.

### DIFF
--- a/xml/common_intro_typografie_i.xml
+++ b/xml/common_intro_typografie_i.xml
@@ -19,53 +19,78 @@
     <dm:translation>no</dm:translation>
     <dm:languages/>
 </dm:docmanager>
-</info>
-<para>
-  The following typographical conventions are used in this manual:
+ </info>
+ <para>
+  The following admonition elements and typographical conventions are used
+  in this documentation:
  </para>
 
  <itemizedlist mark="bullet" spacing="normal">
+  <warning>
+   <para>Vital information to be aware of before proceeding. Warns you about
+    security issues, potential loss of data, damage to hardware, or physical
+    hazards. </para>
+  </warning>
+  <important>
+   <para>Important information to be aware of before proceeding.</para>
+  </important>
+  <note>
+   <para>Additional information, for example about differences in software
+    versions.</para>
+  </note>
+  <tip>
+   <para>A guideline or a piece of practical advice.</para>
+  </tip>
+  <listitem>
+   <screen>&prompt.user;<command>command</command></screen>
+   <para>Commands than can be run by any user, including the &rootuser;
+    user.</para>
+  </listitem>
+  <listitem>
+   <screen>&prompt.root;<command>command</command></screen>
+   <para>Commands that must be run with &rootuser; privileges.
+    Alternatively, you can prefix these commands with the
+    <command>sudo</command> command to run them.</para>
+  </listitem>
+
   <listitem>
    <para>
-    <filename>/etc/passwd</filename>: directory names and file names
-   </para>
+    <filename>/etc/passwd</filename>: directory names and file names </para>
   </listitem>
   <listitem>
    <para>
-    <replaceable>placeholder</replaceable>: replace
-    <replaceable>placeholder</replaceable> with the actual value
-   </para>
+    <replaceable>PLACEHOLDER</replaceable>: replace
+     <replaceable>PLACEHOLDER</replaceable> with the actual value </para>
   </listitem>
   <listitem>
    <para>
-    <envar>PATH</envar>: the environment variable PATH
-   </para>
+    <envar>PATH</envar>: the environment variable PATH </para>
   </listitem>
   <listitem>
    <para>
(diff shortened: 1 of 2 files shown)


xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+   <dm:languages/>
+  </dm:docmanager>
+ </info>
+
+ <para>
+  The following admonition elements and typographical conventions are used
+  in this documentation:
  </para>
 
  <itemizedlist mark="bullet" spacing="normal">
```
tomschr: The `mark` and `spacing` attributes can be removed. They are set to the default values anyway.

xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+  </important>
+  <note>
+   <para>
+    Additional information, for example about differences in software
+    versions.
+   </para>
+  </note>
+  <tip>
+   <para>
+    A guideline or a piece of practical advice.
```
tomschr: What do you think about something like "Optional information"? IMHO your could emphasis the optional nature of a tip more (it can be ignored, it's not essential).

tomschr:
Apart from my comments, looks good to me. :+1: :rocket:

xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+   <dm:status>editing</dm:status>
+   <dm:deadline/>
+   <dm:priority/>
+   <dm:translation>no</dm:translation>
+   <dm:languages/>
+  </dm:docmanager>
+ </info>
+
+ <para>
+  The following admonition elements and typographical conventions are used
```
sknorr: "admonition elements" sounds very much like DocBook/XML lingo (with "admonition" being "Ermahnung"). Maybe "notices" or "warning notices" would be better?

xml/common_intro_typografie_i.xml
sknorr: These are automatically added by the DB 4 to 5 conversion script, I think. Bug? :8ball:

xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+<screen>&prompt.user;<command>command</command></screen>
+   <para>
+    Commands than can be run by any user, including the &rootuser; user.
+   </para>
+  </listitem>
+  <listitem>
+<screen>&prompt.root;<command>command</command></screen>
+   <para>
+    Commands that must be run with &rootuser; privileges. Alternatively, you
+    can prefix these commands with the <command>sudo</command> command to
```
sknorr: "...Alternatively, in many cases you can..."
Not sure if that ruins the sentence, though. [I do not have a good example where sudo does not work on SUSE at the moment. But I think I have seen that multiple times.]

xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+  </warning>
+  <important>
+   <para>
+    Important information to be aware of before proceeding.
+   </para>
+  </important>
+  <note>
+   <para>
+    Additional information, for example about differences in software
+    versions.
```
sknorr: Already much better than the style guide version. Thanks.

xml/common_intro_typografie_i.xml
```diff
@@ -11,20 +11,59 @@
...
+
+ <para>
+  The following admonition elements and typographical conventions are used
+  in this documentation:
  </para>
 
  <itemizedlist mark="bullet" spacing="normal">
+  <warning>
+   <para>
+    Vital information to be aware of before proceeding. Warns you about
```
sknorr: "to be" => "you should be"? A bit longer but more personal imho.

xml/common_intro_typografie_i.xml
sknorr: Not sure about "optional information": it is not like you could opt in to to something here. I had the words "helpful" and "background information" in my mind but could not find a way to express where I'd like to go here.

xml/common_intro_typografie_i.xml
taroth21: Changed this to: In many cases you can also prefix [...]

xml/common_intro_typografie_i.xml
taroth21: For the <warning/>, I have changed this to: Vital information you must be aware of before proceeding. For <important/>, I would use: Important information you should be aware of before proceeding.

xml/common_intro_typografie_i.xml
taroth21: What about 'Additional or background information'? But I think I'd rather prefer to stick to 'additional information'.

xml/common_intro_typografie_i.xml
taroth21: Good point, changed it to 'notices' now (because not all of them are for warning purposes).

PR #10, opened 2016-02-03
Feature/crowbar ui

taroth21:
Developers: Please search the file 'xml/depl_crowbar.xml' for my remarks containing the keyword 'DEVs' and answer the questions/provide some input. Many thanks!


xml/depl_crowbar.xml
```diff
@@ -0,0 +1,322 @@
...
+  
+  <procedure xml:id="pro.depl.crow.login">
+   <title>Logging In to the &crow; Web Interface</title>
+   <step>
+    <para> On any machine, start a Web browser and make sure that JavaScript and
+     cookies are enabled.</para>
+   </step>
+   <step>
+    <para> As URL, enter the IP address of the &admserv;, for example:</para>
+    <screen>http://192.168.124.10:3000/</screen>
```
MaximilianMeister: I think we can drop the port here. It runs on port 80 by default now.

xml/depl_crowbar.xml
```diff
@@ -0,0 +1,322 @@
...
+       <remark>taroth 2016-02-03: DEVs, can I open the resulting report from
+        here? I tried to run a report but it seems to run forever with no result (see
+        bsc#964946)</remark></para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term><guimenu>Backup &amp; Restore</guimenu></term>
+     <listitem>
+      <para>This screen lets you create a backup of the &admserv; and download
+       it. You can also restore a backup or upload a backup image from your file
```
MaximilianMeister: MINOR: maybe `You can also restore _from_ a backup` ?
and maybe `upload from your _local_ filesystem`

taroth21: Yep, much better! Thanks!

xml/depl_crowbar.xml
taroth21: Changed, thanks!

xml/depl_crowbar.xml
```diff
@@ -0,0 +1,327 @@
...
+       is marked red in the screen. Any repositories marked green are usable and
+       available to each node in the cloud. If you activate the check box for a
+       repository in the <guimenu>Active</guimenu> column, the managed nodes
+       will automatically be configured to use this repository. <remark>taroth
+        2016-02-05: DEVs: and if the check box is *not* ticked? will the nodes
+        use another repo? or no repo at all? and what's the default - no repo
+        activated?</remark></para>
+      <para>You cannot edit any repositories in this screen. The list of
+       repositories shown here is defined in
+        <filename>/etc/crowbar/repos.yml</filename>. If you need additional,
```
MaximilianMeister:
```
The list of repositories shown here is defined in <filename>/etc/crowbar/repos.yml</filename>
```
is not correct actually. the list is defined in `/opt/dell/crowbar_framework/config/repos-[cloud|ses].yml` but that is not important for the user.

I would`just write
```
The list of repositories shown here can be extended by adding other/third-party repositories in <filename>/etc/crowbar/repos.yml</filename>
```

we should maybe provide a hint in the documentation of how `/etc/crowbar/repos.yml` has to look.
@vuntz do you think that makes sense, and what would you add as an example repo?

here is a minimal example from what mkcloud adds.
```yml
---
suse-12.1:
  x86_64:
    SLES12-SP1-Updates-test:
      url: 'http://dist.nue.suse.com/ibs/SUSE:/Maintenance:/Test:/SLE-SERVER:/12-SP1:/x86_64/update/'
```

vuntz: Yes, having an example sounds good. On top of the url attribute, you can have a priority attribute (that sets the repo priority for zypper) and a ask_on_error attribute (that sets the ask_on_error flag in the autoyast profile for that repo).

taroth21: @MaximilianMeister: Thanks! I rephrased the sentence about /etc/crowbar/repos.yml. However, I would not want to use the example from mkcloud since it points to an internal URL.  
@vuntz, @MaximilianMeister: In the test deployment I'm looking at; i found the following:
```suse-12.1:
  x86_64:
    sles12-sp1-pool:
      name: "SLES12-SP1-Pool"
      required: "mandatory"
      features: ["os"]
      repomd:
        tag: "obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/SLES/12.1/POOL/x86_64"
        key_md5: "9a62177e1c6852d48453ed3909956d6b"
      url:
      smt_path: "SUSE/Products/SLE-SERVER/12-SP1/x86_64/product"
      ask_on_error: false
```
Please let me know if we can use that one as an example (it's taken from /opt/dell/crowbar_framework/config/repos-cloud.yml, however) or what I need to remove for a minimal example (only containing the minimum required settings).

MaximilianMeister: @taroth21 just use

```yaml
---
suse-12.1:
  x86_64:
    Custom-Repo-12.1:
      url: 'http://example.com/12-SP1:/x86_64/custom-repo/'
      ask_on_error: true # sets the ask_on_error flag in the autoyast profile for that repo
      priority: 99 # sets the repo priority for zypper
```

@vuntz okay like that?

xml/depl_crowbar.xml
```diff
@@ -0,0 +1,368 @@
...
+      <para>This is the default view after logging in to the &crow;
+       &wi;. The &dash; shows the groups (which you can create to
+       arrange nodes according to their purpose), which nodes belong to each
+       group and which state the nodes and groups are in. In addition, the
+       total number of nodes is displayed in the top-level row.</para>
+      <para>The color of the dot in front of each node indicates the node
+       status:</para>
+      <itemizedlist>
+       <listitem>
+        <para>Gray means the node is being discovered by the
```
toabctl: This is imo a bit missleading because you say "gray means discovered" and in the next listitem, "yellow means discovered"

taroth21: Well, if you read more closely, it says "Gray means the node *is being* discovered" whereas "Yellow means the node *has been* discovered", which is a difference in both tense and meaning.

vuntz: Please note that gray is also used when there's no up-to-date information about a deployed node (which typically means that chef-client is not running anymore on the node).

taroth21: Thanks! I added this information.

PR #15, opened 2016-02-18
added information from bsc#956900, two questions still open

taroth21:
@mjura: Please review the changes and let me know what to enter for 'Datastore Names' (see the FIXME in xml/depl_vmware.xml). Thanks!

### DIFF
--- a/xml/depl_vmware.xml
+++ b/xml/depl_vmware.xml
@@ -174,11 +174,15 @@
     </listitem>
    </varlistentry>
    <varlistentry>
-    <term><guimenu>Datastore Names</guimenu></term>
+    <term><guimenu>Regex to match the name of a datastore</guimenu></term>
     <listitem>
-     <para>
-     FIXME
-     </para>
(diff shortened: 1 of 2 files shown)


xml/depl_vmware.xml
```diff
@@ -174,7 +174,15 @@
     </listitem>
    </varlistentry>
    <varlistentry>
-    <term>VLAN Interface</term>
+    <term><guimenu>Datastore Names</guimenu></term>
```
mjura: Regex to match the name of a datastore.

This setting specifies the data stores to use with Nova Compute. For example, value  "nas.*" selects all the data stores that have a name starting with "nas". If this line is omitted, Nova Compute uses the first data store returned by the vSphere API. It is recommended not to use this field and instead remove data stores that are not intended for OpenStack.

mjura:
Looks OK, except `Datastore Name`

xml/depl_vmware.xml
taroth21: Thanks, added that.

PR #20, opened 2016-03-18
improvements to cluster recovery docs (bsc#966643)

aspiers:
/cc @vuntz

### DIFF
--- a/xml/depl_ha_recovery.xml
+++ b/xml/depl_ha_recovery.xml
@@ -140,18 +140,6 @@
     <para>Re-register the node with &crow; by executing:</para>
     <screen>&prompt.root;service crowbar_join start</screen>
    </step>
-   <!--
-   taroth 2016-02-24: aspiers, is the following still (or also?) needed?
-   copied from depl_nodex.xml, l.625-->
-   <!--
-    <listitem>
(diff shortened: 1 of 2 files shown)


xml/depl_ha_recovery.xml
```diff
@@ -130,47 +138,40 @@
...
-    <listitem>
-    <para>
-     On the &admserv;, run the following command to update the status
-     of the node specified with <replaceable>NODE</replaceable>.
-    </para>
-    <screen>crowbar crowbar transition <replaceable>NODE</replaceable> ready</screen>
-   </listitem>
-   -->
    <step>
     <para>Log in to one of the <emphasis>other</emphasis>
```
aspiers: IIRC, this step was to ensure `haproxy.cfg` is correct (Crowbar might have dropped nodes from it if they were down for long enough), in which case it needs to be done for *all* other nodes in the cluster.  Right @vuntz?

vuntz: Without reading the whole thing, yes, I think you're correct. Yet another example of how broken things are with how we deal with the run lists ;-)

taroth21:
@aspiers: Thanks!

PR #22, opened 2016-04-18
Fix nova barclamp paths for Cloud 5

bhdn:
It was introduced by dd4ad5129e4a16b0ea82fcd0dd6c7fac4b266d76


taroth21:
@bhdn: Tanks!

PR #24, opened 2016-07-27
make interface parameter names match list

aspiers:
Each component in physical interface definitions (e.g. "10g2") should
the the parameter name correspond to the subtitle of its long
description in the list below it, otherwise it's confusing.


fsundermeyer:
Thanks!

PR #26, opened 2016-11-01
add caveats about HA config sharing networks

aspiers:


aspiers:
@vuntz Do we want to declare certain network configs as unsupported? This could be tricky if existing customers are already using them though.

rhafer:
@aspiers Hm, if I understand this correctly it means that a sensible HA setup (with sbd on iscsi) needs at least 6 physical interfaces. 2 for corosync/admin, 2 for sbd and 2 for sdn. Or am I missing something?

aspiers:
@rhafer Not quite; IIUC the SBD link does not require bonding.  According to the `sbd(8)` man page:

    If the Pacemaker integration is activated, "sbd" will not self-fence if 
    device majority is lost, if:

    1.  The partition the node is in is still quorate according to the CIB;
    2.  it is still quorate according to Corosync's node count;
    3.  the node itself is considered online and healthy by Pacemaker.

So I think 5 is the minimum, or even 4 if you use a single NIC for sdn - but in that case, you better also be using compute node HA ;-)

However we should check all this with the HA gurus.

rhafer:
> So I think 5 is the minimum, or even 4 if you use a single NIC for sdn

Hm, I wonder how that will be received in reality at the customers :). Since we also usually recommend yet another set of interfaces for storage ...

aspiers:
Conversation just now with @l-mb and @krig on `#sleha` actually contradicts this - apparently it's OK if SBD is on the same bonded interface as corosync.  If that interface loses connectivity, SBD will self-fence which is the correct behaviour.  So I'm dropping the second paragraph about SBD needing to be kept separate to corosync.

aspiers:
Adjusted the SBD paragraph based on `#sleha` discussions.  @l-mb, @krig: your review on this would be most welcome ;-)

krig:
Tentative thumbs up on the current formulation, though really I am just parroting @l-mb on this (unless I say something that is incorrect, the mistakes are my own).

aspiers:
Expanded the commit message in order to practice what I preach.

aspiers:
@l-mb said "looks OK to me" on IRC.

xml/depl_require.xml
```diff
@@ -2069,6 +2069,24 @@
         between the cluster nodes) to ensure redundancy.
        </para>
       </important>
+      <important>
+       <title>Dedicated networks</title>
+       <para>
+        The <literal>corosync</literal> network communication layer is
```
taroth21: My proposal (sentence length, clarity, plus trying to integrate feedback by @roseswe on https://github.com/SUSE-Cloud/doc-cloud/pull/30):
[...] critical to the health of a cluster. Therefore, use a dedicated bond for both Admin network and corosync communication and a second bond for the OpenStack traffic.

xml/depl_require.xml
```diff
@@ -2069,6 +2069,24 @@
         between the cluster nodes) to ensure redundancy.
        </para>
       </important>
+      <important>
+       <title>Dedicated networks</title>
```
taroth21: Please use title style capitalization (http://docserv.nue.suse.com/documents/Styleguide/docu_styleguide/html/index.html#sec.capitalization): Dedicated networks -> Dedicated Networks

taroth21:
@aspiers: If your time allows, could you have a look at my comments during the next days and let me know how to proceed? Thanks!

xml/depl_require.xml
aspiers: Hmm not quite - that's too prescriptive.  The OpenStack traffic could be spread over multiple interfaces, some or all of which are bonded.

aspiers: But you helped me realise that actually we *force* corosync traffic to go over the admin network.  So I'll reword it to make that clearer.

xml/depl_require.xml
aspiers: Done.

xml/depl_require.xml
```diff
@@ -2069,6 +2069,24 @@
...
+        The <literal>corosync</literal> network communication layer is
+        critical to the health of the cluster, so it should not only be
+        hosted by redundant communication paths, but should also avoid
+        being placed on interfaces shared with any other networks which
+        could experience heavy load, such as the public / private / SDN /
+        storage networks.
+       </para>
+       <para>
+        Similarly, if SBD over iSCSI is used as a STONITH device (see
+        <xref linkend="vle.ha.req.stonith"/>) then it should also ideally
```
aspiers: Reworded.

aspiers:
@taroth21 Hopefully your concerns are all addressed now?

taroth21:
Thanks @aspiers! Because depl_require.xml in 'develop' had changed this morning, I have merged your contents into the updated file. I also tried to simplify some sentences for the sake of non-native English speakers/readers. You can find my changes in https://github.com/SUSE-Cloud/doc-cloud/commit/35d23118. In case you are not OK with the changes, please let me know!  Thx!

aspiers:
@taroth21 Why are these changes not visible in http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/ ?  Is that intentional?

taroth21:
@aspiers: That's a very good question indeed! I'm not sure what happened here, but the changes also do not show up in the source code - really sorry for that and thanks for spotting this! I re-added the changes with https://github.com/SUSE-Cloud/doc-cloud/commit/1bdb9098d34ce4fa4737e6f477c1c922c53d3ff9.

taroth21:
... and removed the unneeded brackets in the first sentence with https://github.com/SUSE-Cloud/doc-cloud/commit/3fb5df61. The changes are now visible in http://docserv.suse.de/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.reg.ha.general

aspiers:
@taroth21 Thanks! Looks like it got accidentally reverted by cf8303b39.

PR #29, opened 2016-12-09
fix wording 'teaming mode'/'network device bonding'

taroth21:
- as suggested by @dirkmueller  to avoid confusion
  with 'bonding policy'


taroth21:
@dirkmueller: ping?

PR #30, opened 2016-12-12
Bugfix/bsc 960523

taroth21:
@aspiers, @roseswe: Please review, thanks!

### DIFF
--- a/xml/depl_require.xml
+++ b/xml/depl_require.xml
@@ -2105,7 +2105,7 @@
        <para>
         For a supported &hasetup;, it is required to set up cluster
         communication via two or more redundant paths. For this purpose, use
-        teaming network mode in your network setup. For details, see
+        network device bonding and teaming network mode in your &crow; network setup. For details, see
         <xref
          linkend="sec.depl.req.network.modes.teaming"/>. At
         least two Ethernet cards per cluster node are required for network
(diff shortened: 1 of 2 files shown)


taroth21:
@dirkmueller : thanks! @aspiers : are you fine with this, too?

xml/depl_require.xml
```diff
@@ -2116,10 +2116,14 @@
...
-       Bonding</citetitle>, and <citetitle>Installation and Basic
-       Setup</citetitle> (section <citetitle>Defining the Communication
-       Channels</citetitle>).
+       read the following chapter: <citetitle>Network Device
+       Bonding</citetitle>.
+      </para>
+      <para>
+       Using a second communication channel (ring) in &corosync; (as an
+       alternative to network device bonding) is not supported yet in &productname;.
+       By default, &productname; uses <literal>eth0</literal> (admin network) for
```
aspiers: Do we explicitly want to mention `eth0`?  Don't some servers end up with interfaces like `em0` or other weird and wonderful names?  And anyway the admin network could be on any interface in theory.  I would have thought it would make more sense to only mention the admin network, or at least word it as:

    By default, &productname; uses the admin network (typically <literal>eth0</literal>)

bmwiedemann:
Am 04.01.2017 um 19:12 schrieb Adam Spiers:
> Do we explicitly want to mention |eth0|? Don't some servers end up with
> interfaces like |em0|

indeed on Dell machines first interface is em1 via biosdevname magic

taroth21:
@dirkmueller , @aspiers: Thank you for the review and the comments! I have integrated Adam's proposal and will merge the PR now.

PR #33, opened 2016-12-29
Feature/openstack cloud 7 barclamp ui changes

dmpop:
Developers, please review the changes and provide the necessary info. Search for DEVS FIXME. Thank you!


xml/depl_nodes.xml
```diff
@@ -996,14 +996,36 @@ getent group postgres | cut -d: -f3</screen>
...
+   </varlistentry>
+   <varlistentry>
+    <term>
+     <guimenu>
+      Default Credentials: Create Regular User
+     </guimenu>
+    </term>
+    <listitem>
+     <para>
+      Enable or disable the ability to create the regular user.
```
dirkmueller: its not whether you have the *ability* to create a regular user, it is about whether a default user (with a default password) is created *automatically* (which you normally don't want, especially in a LDAP environment where you're operating on possibly read only identity information).

xml/depl_nodes.xml
```diff
@@ -996,14 +996,36 @@ getent group postgres | cut -d: -f3</screen>
...
+   </varlistentry>
+   <varlistentry>
+    <term>
+     <guimenu>
+      Default Credentials: Regular User Username/Password
+     </guimenu>
+    </term>
+    <listitem>
+     <para>
+      User name and password for the regular user. Both the regular user and the administrator accounts can be used to log in to the &cloud; &dash; to manage &o_ident; users and access.
```
dirkmueller: obviously the regular user can *not* manage users..

xml/depl_nodes.xml
```diff
@@ -3195,6 +3195,30 @@ getent group postgres | cut -d: -f3</screen>
...
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term><guimenu>
+      Scheduler Options: Reserved Memory for Nova Compute hosts (MB)
+     </guimenu>
+    </term>
+    <listitem>
+     <para>
+      Amount of memory reserved for Nova Compute hosts.<remark condition="accuracy">2017-01-02 - dpopov: DEVS FIXME Check whether this is correct.</remark>
```
dirkmueller: its the reverse, amount of Host memory reserved and not used for allocating VMs by nova-compute

xml/depl_nodes.xml
```diff
@@ -4223,9 +4223,10 @@ getent group postgres | cut -d: -f3</screen>
 
   <variablelist>
     <varlistentry>
-      <term></term>
+      <term>Use Ceph deployed by Crowbar</term>
       <listitem>
 	<para>
+	  Set to <systemitem>true</systemitem> to use Ceph deployed with Crowbar.
```
kbaikov: Can native English speakers suggest the correct preposition (by/with).
And make it consistent? @aspiers maybe?

xml/depl_nodes.xml
dirkmueller: yes

xml/depl_nodes.xml
```diff
@@ -2275,6 +2315,92 @@ getent group postgres | cut -d: -f3</screen>
...
 
+  <bridgehead renderas="sect2"><guimenu>Hitachi HUSVM</guimenu>
+  </bridgehead>
+
+  <variablelist>
+    <varlistentry>
+      <term>Storage Protocol</term>
+      <listitem>
+      <para>
+	Choose between the FibreChannel and iSCSI protocols.<remark>2016-12-29 - dpopov: DEVS FIXME Please check whether this is correct.</remark>
```
dirkmueller: LGTM

xml/depl_nodes.xml
```diff
@@ -2275,6 +2315,92 @@ getent group postgres | cut -d: -f3</screen>
...
+      <para>
+	None<remark>2016-12-29 - dpopov: DEVS FIXME Please provide info.</remark>
+      </para>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term>Serial number of storage system</term>
+      <listitem>
+      <para>
+	None<remark>2016-12-29 - dpopov: DEVS FIXME Please provide info.</remark>
```
dirkmueller: this comes from http://docs.openstack.org/newton/config-reference/block-storage/drivers/hitachi-storage-volume-driver.html which doesn't provide more information. My suggestion would be to link the upstream config reference instead of trying to come up with better documentation here. this applies to all hitachi options.  We can much quicker fix that upstream than in our documentation, and then it benefits everyone.

xml/depl_nodes.xml
```diff
@@ -2809,6 +2905,56 @@ getent group postgres | cut -d: -f3</screen>
...
+	<para>
+	  xCAT login credentials.
+	</para>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term>rdev list for physnet1 vswitch uplink (if available)</term>
+      <listitem>
+	<para>
+	  RDEV address of the OSA cards which are connected to the vswitch.<remark>
```
dirkmueller: list of rdev addresses that should be connected to this vswitch

xml/depl_nodes.xml
```diff
@@ -2809,6 +2905,56 @@ getent group postgres | cut -d: -f3</screen>
...
+	    2016-12-29 - dpopov: DEVS FIXME Please check whether this is correct.</remark>
+	</para>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term>xCAT IP Address on Management Network</term>
+      <listitem>
+	<para>
+	  IP address of the xCAT management interface.<remark>
+	    2016-12-29 - dpopov: DEVS FIXME Please check whether this is correct.</remark>
```
dirkmueller: :+1:

xml/depl_nodes.xml
dmpop: Makes sense. Will do.

PR #38, opened 2017-01-30
Fix paragraph about admin node requirements

rsalevsky:
The paragraph about the dynamic RAM requirements belongs to the admin
server.

@fsundermeyer This was probably my fault, sorry for the inconvenience.

PR #39, opened 2017-01-31
Rework RAM requirements

vuntz:
We need more for the controllers, and the part that depends on the
number of nodes is actually for the admin server.


rsalevsky:
Partly duplicates #38 but this is also fine with me.

vuntz:
> Partly duplicates #38 but this is also fine with me.

Oh, sorry, hadn't seen that.

PR #41, opened 2017-02-08
clarify why compute HA needs shared storage for instances

aspiers:
What do you guys think about this?  Does this shared storage consideration actually belong in [the Considerations and Requirements section of the Architecture and Requirements chapter](http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.reg.ha.compute) instead?

### DIFF
--- a/xml/depl_nodes.xml
+++ b/xml/depl_nodes.xml
@@ -3572,10 +3572,16 @@ getent group postgres | cut -d: -f3</screen>
     <title>Shared Storage</title>
     <para>
      It is recommended to use shared storage for the
-     <filename>/var/lib/nova/instances</filename> directory. If an external NFS
-     server is used, enable the following option in the &o_comp; &barcl;
-     proposal: <guimenu>Shared Storage for Nova instances has been manually
-      configured</guimenu>.
+     <filename>/var/lib/nova/instances</filename> directory, in order
+     to ensure that ephemeral disks will be preserved during recovery
+     of VMs from failed compute nodes.  Without shared storage, any
+     ephemeral disks will be lost, and recovery will rebuild the VM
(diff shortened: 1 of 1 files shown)


cjschroder:
The clarification is good 👍 There is already a section on Persistent Storage in the Considerations and Requirements chapter, http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.req.storage

aspiers:
@cjschroder OK thanks - I guess anyone from your side is free to merge when you deem appropriate :)

taroth21:
@aspiers , @cjschroder: Thanks Adam and Carla! I also like the idea of mentioning the general part of this tip in the Considerations and Requirements chapter (however, I would keep the last sentence about the barclamp in depl_nodes.xml as it is rather specific). At the moment, I'm not sure where it would fit best because  http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.req.storage is mostly about persistent storage (IMHO not a good choice) but contains a pointer to http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.req.storage.hardware.compute which talks about both persistent and ephemeral storage (IMHO, a better choice). And then we also have a section about HA requirements (http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/single-html/#sec.depl.req.ha), which might be an option, too. Your ideas?

cjschroder:
@taroth21 I'll look for suitable places to reference this as I re-org the appendixes.

PR #48, opened 2017-03-29
Feature/gui upgrade

dmpop:
I would be grateful if you could take a look at the Upgrading Using the Web Interface chapter and see whether I got everything right. Thanks!


xml/depl_maintenance.xml
```diff
@@ -391,17 +391,133 @@
...
+      cannot be higher than the number of available disks. Check the
+      <guimenu>Replica</guimenu> setting in the &o_objstore; &barcl; prior to
+      starting the upgrade. If it is higher than the number of available disks,
+      adjust it to match the number of disks and redeploy the &barcl;.
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      Upgrading is not supported if &o_blockstore; has been deployed with the
+      <literal>raw devices</literal> back-end.
```
jsuchome: This is not exactly true. For **non-disruptive upgrade** cinder backend cannot be raw devices or local file. Normal upgrade can use them

xml/depl_maintenance.xml
```diff
@@ -419,14 +535,239 @@
       &compnode;s needs to be sufficient to evacuate the nodes one by one.
      </para>
     </listitem>
+    <listitem>
+     <para>
+      &o_img; needs to be set up with shared storage; that means the
+      <guimenu>Default Storage Store</guimenu> value in the &o_img; must
+      <emphasis>not</emphasis> be set to <literal>File</literal>.
```
jsuchome: This is not exactly true. I've pointed in the bug report to http://docserv.nue.suse.com/documents/Cloud7/suse-openstack-cloud-deployment/html/cha.depl.ostack.html#sec.depl.ostack.glance.ha

which says (I paraprase): **if** there's glance store set up to File, it should be configured as a shared storage.

dmpop: Which bug report are you referring to?

jsuchome: The on in commit message that added this text :-) bsc#1030676

dmpop: Ah! :-)

dmpop: I'm a bit confused. The bug you are referring to is closed in https://github.com/SUSE-Cloud/doc-cloud/commit/05836e21 with the exact same wording as in this document.

jsuchome: Yes, and that is wrong. I did not close the bug

dmpop: Got it! So is the following correct?

*If the Default Storage Store value in Glance is set to File, then Glance must be configured as a shared storage.*

jsuchome: That sounds correct. Well you might add that it is for non-disruptive upgrade to work properly (for normal case we do not care).

dmpop: Will do! Thanks!

dmpop:
Both change requests have been implemented. Please approve. :-)

xml/depl_maintenance.xml
```diff
@@ -448,8 +448,10 @@
     </listitem>
     <listitem>
      <para>
-      Upgrading is not supported if &o_blockstore; has been deployed with the
-      <literal>raw devices</literal> back-end.
+       Note that a non-disruptive upgrade is not supported if the
+       &o_blockstore; has been deployed with the <literal>raw devices</literal>
+       or <literal>local file</literal> back-end. In this case, you have to
+       perform a regular upgrade.
```
jsuchome: User either has to perform 'normal' upgrade, or there might be a chance to change the cinder backend (just like with other checks for possibility of non-disruptive upgrade)

dmpop: Fixed in fb10c02

xml/depl_maintenance.xml
```diff
@@ -559,7 +559,7 @@
      <informalfigure>
       <mediaobject>
        <imageobject role="fo">
-        <imagedata fileref="depl_upgrade_initial.png" width="100%" format="png"/>
+        <imagedata fileref="depl_upgrade_initial.png" width="75%" format="png"/>
```
fsundermeyer: The width for fo should be 100% for all screenshots here (or, 99% if that does not work).

Sorry to be a pain, but the screenshots are a liitle too wide and on most of them there is a lot of unused gray space. Can you try to shoot them with a height that fits the content and less width (by choosing the minimal browser width that is possible?). The bigger the screenshots are, the harder it is to read them in PDF...

dmpop: No problem. Will do!

dmpop: Should be fixed now.

dmpop:
Depends on how you read it: 1) the output (+ error message) offers, ") *both* output and error message offer. I fixed it to avoid ambiquity.

jsuchome:
@dmpop thanks... btw, where could I see latest state of documentation (not just XML diff)?

dmpop:
Nightly builds are generated from the develop branches and are available at http://docserv.suse.de/documents/ But the stuff I'm working on right now has not yet been merged into the develop branch, so you can only see the raw XML in the feature/gui_upgrade branch.

dmpop:
@jsuchome It would be great if you could take a quick look at b2e27fd1a5fa28a838af16460a2b7e31c5e793ca Thanks!

xml/depl_maintenance.xml
```diff
@@ -565,10 +565,14 @@
...
     <listitem>
      <para>
-       If your &productname; 6 setup is configured to use LBaaS v1, you need to
+       If your &productname; 6 setup is using LBaaS v1, you need to
        switch to LBaaS v2 manually. To do this, switch to the
        <systemitem>Raw</systemitem> view in the &o_netw; &barcl; and set the
-       <systemitem>use_lbaasv2</systemitem> parameter to <literal>true</literal>.
+       <systemitem>use_lbaasv2</systemitem> parameter to
+       <literal>true</literal>. It is also necessary to re-create all the
+       object manually. This includes load balancers pools, and health
```
jsuchome: *load balancers, pools*

Just note that users know what they should do: they've created those objects manually before, otherwise there wouldn't be any.

dmpop: how about the following?

*You also must re-create all objects manually following the procedure that was used to create them.*

jsuchome: I think, just leave as it was, just add a comma, as pool is something different than load balancer.

(And don't ask me what those are, I have no idea)

dmpop: Got it!

> (And don't ask me what those are, I have no idea)

I know the answer to this question: they are **objects**.  :-)

Thanks!

PR #49, opened 2017-03-29
A bit of heavier editing. Please make sure no

cjschroder:
meanings are changed, e.g. changing "The latter supports several back-ends, including Ceph, that are deployed during the installation." from "...that can be deployed during the installation."


fsundermeyer:
Thanks a lot, Carla, nice work!

PR #53, opened 2017-04-10
Light edits section 2.6 thru 2.8

cjschroder:
It is a pleasure to read these manuals. Very nice work.


taroth21:
Very good catches, many thanks!

PR #57, opened 2017-04-13
ch. 5 proofread & edit.

cjschroder:


xml/depl_conf_admin_repos.xml
```diff
@@ -66,22 +65,20 @@
    <tip>
     <title>Providing the &productname; Repository via HTTP</title>
     <para>
-     The &sls; product repositories needs to be made available locally
-     to enable booting using PXE for node deployment. The &productname;
+     The &sls; product repositories need to be available locally
+     to enable PXE booting for node deployment. The &productname;
```
fsundermeyer: According to the style guide, we mus not use "PXE boot", but rather "(to) boot using PXE or (to) boot via PXE"

cjschroder: allrightythen, fixed. Thank you!

PR #63, opened 2017-04-18
Add SLEShammer to the glossary

cjschroder:
I also added the subtitle 'A glossary of terms and product names' so that searches for 'glossary' would find it, and make it more prominent in the TOC

### DIFF
--- a/xml/glossary.xml
+++ b/xml/glossary.xml
@@ -10,7 +10,7 @@
  the other terms it might make more sense to include the upstream glossary that is appended to all
  upstream manuals -->
 <glossary xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0" xml:id="gl.cloud">
-    <title>Terminology: A glossary of terms and product names</title>
+    <title>Glossary of Terminology and Product Names</title>
      <info>
  <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
     <dm:maintainer>fs</dm:maintainer>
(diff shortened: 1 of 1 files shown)


xml/glossary.xml
```diff
@@ -10,9 +10,9 @@
  the other terms it might make more sense to include the upstream glossary that is appended to all
  upstream manuals -->
 <glossary xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0" xml:id="gl.cloud">
- <title>Terminology</title>
- <info>
-<dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
+    <title>Terminology: A glossary of terms and product names</title>
```
taroth21: Good idea! But I think the subtitle needs title style  to match the style guide.

xml/glossary.xml
```diff
@@ -962,6 +962,13 @@
    </para>
   </glossdef>
  </glossentry>
+  <glossentry xml:id="gloss.hammer"><glossterm>SLEShammer</glossterm>
+  <glossdef>
+   <para>
+    When you first boot a node in Cloud via PXE, it is booted with the SLEShammer image. This performs the initial hardware discovery, and  registers the node with crowbar. Once you allocate the node, it is rebooted with a regular SLES installation image.
```
taroth21: Please use entities: When you first boot a node in &productname; via PXE, it is booted with the SLEShammer image. This performs the initial hardware discovery, and registers the node with &crow;.

taroth21: Once you allocate the node [...] Not sure about using "once" here as the next style checker run will probably recommend to replace it with "after".

xml/glossary.xml
cjschroder: I should have said "longer title". I added a subtitle and it did not appear in the TOC, so I made a longer title instead.

xml/glossary.xml
cjschroder: Thank you, fixes applied, title amended.

taroth21:
Thanks!

PR #65, opened 2017-04-19
Ch 11-11.1 Light edits & proofread

cjschroder:


xml/depl_nodes.xml
```diff
@@ -552,14 +549,14 @@
     </listitem>
    </varlistentry>
    <varlistentry>
-    <term><guimenu>HAProxy: Public name for public virtual IP</guimenu>
+    <term><guimenu>HAProxy: Public name for public virtual IP address</guimenu>
```
taroth21: Hm, as this is within a guimenu element, the string is called like that in the SW. Unless we can convince the developers to change the string, we should keep it (at least, within the guimenu element).

cjschroder: Yes, thanks, it must be exactly the same.

taroth21:
Many thanks!

PR #66, opened 2017-04-19
ch 11.2-11.4 proofread & light edits

cjschroder:


xml/depl_nodes.xml
```diff
@@ -1086,7 +1082,7 @@ getent group postgres | cut -d: -f3</screen>
      </para>
      <variablelist>
       <varlistentry>
-       <term><guimenu>Generate (self-signed) certificates</guimenu>
```
taroth21: Better to keep the text within the guimenu tags as the SW string is called like that, otherwise manual and SW do not match. Alternatively, we need to file a bug report to change the SW string.

cjschroder: Of course, good catch.

taroth21:
Thanks!

PR #67, opened 2017-04-19
Ch 11.5-11.7 many small corrections

cjschroder:
Please make sure changes to XML tags are correct.

### DIFF
--- a/xml/depl_nodes.xml
+++ b/xml/depl_nodes.xml
@@ -1306,15 +1306,14 @@ getent group postgres | cut -d: -f3</screen>
   <title>Deploying &ceph; (optional)</title>
 
   <para>
-   &ceph; adds a redundant block storage service to &cloud;. It lets
-   you store persistent devices that can be mounted from &vmguest;s. It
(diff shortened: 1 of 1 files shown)


xml/depl_nodes.xml
```diff
@@ -1773,8 +1764,8 @@ getent group postgres | cut -d: -f3</screen>
     </term>
     <listitem>
      <para>
-      Enables the possibility to extract tar files into a swift account and
-      to delete multiple objects or containers with a single request. See
+      Extract tar files into a Swift account, and
```
taroth21: Stefan just suggested to replace 'tar file' with 'TAR archive' because they are synonyms anyway...

cjschroder: tar = tape archive, so it's a tape archive archive \o/  Old UNIX terms are fun.

taroth21:
Thanks, Carla. :)

xml/depl_nodes.xml
taroth21: Hehe, maybe we should ask Stefan to reconsider. ;)

PR #70, opened 2017-05-03
Replace lbaasv1 commands with lbaasv2 commands in admin guide

cjschroder:
re bug #1031679. soc 7 uses openstac newton  which supports only lbaasv2
refs: https://media.readthedocs.org/pdf/f5-openstack-lbaasv2-driver/master/f5-openstack-lbaasv2-driver.pdf
https://docs.openstack.org/newton/networking-guide/config-lbaas.html


taroth21:
Thanks!

PR #74, opened 2017-05-11
Feature/monasca barclamp

dmpop:
Deploying Monasca


xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+	  communicating with &o_ident;. If set to <literal>false</literal>, the
+	  <literal>ca_file</literal> option must be disabled as well.
+        </para>
+      </listitem>
+      </varlistentry>
+      <varlistentry>
+      <term>ca_file</term>
+      <listitem>
+        <para>
+          Specifies whether a CA certificate is used for verifying &o_ident;'s SSL certificate.
```
jgrassler: This doesn't specify whether to use a CA certificate for verification (that is what the `insecure` setting is for) but the filesystem location of such a CA file.

xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+	  <literal>http_check</literal>) that are allowed run simultaneously. The default
+	  value <literal>1</literal> means that plug-ins are run sequentially.
+        </para>
+      </listitem>
+      </varlistentry>
+      <varlistentry>
+      <term>pool_full_max_retries</term>
+      <listitem>
+        <para>
+	  If a problem with multiple plug-ins results is blocking the entire
```
jgrassler: nit: s/plug-ins/plug-ins'/

xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+        </para>
+      </listitem>
+      </varlistentry>
+      <varlistentry>
+      <term>amplifier</term>
+      <listitem>
+        <para>
+	  Number of extra dimensions to add to metrics sent to the
+	  &o_monitor; API. This option is intended for load testing
+	  purposes only. Do not enable the option in production! The default <literal>0</literal> value disables the option.
```
jgrassler: Maybe "disables the addition of dimensions" rather than "disables the option"?

xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+      <varlistentry>
+      <term>repo_dir</term>
+      <listitem>
+        <para>
+	  List of directories for storing elasticsearch snapshots. Must be
+	  created manually and be writeable by the <systemitem
+	  class="username">elasticsearch</systemitem> user. 
+	  <!-- dpopov 2017-05-10:
+     FIXME: (Add reference the correct book/chapter) See the CMM operator's guide, section 3.4.1 for instructions on
+     creating an elasticsearch snapshot. --> Must contain at least one entry.
```
jgrassler: No, it needn't contain any entries. By default it's empty, which means that there is no snapshot directory and snapshotting will not work.

xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+<bridgehead renderas="sect3"><guimenu>master: configuration for
+monasca-installer on the &crow; node</guimenu></bridgehead>
+
+<variablelist>
+  <varlistentry>
+      <term>influxdb_retention_policy</term>
+      <listitem>
+        <para>
+	 Number of days to keep metrics records in influxdb. Refer to <link
+	 xlink:href="https://docs.influxdata.com/influxdb/v0.9/query_language/database_management/#create-retention-policies-with-create-retention-policy"/>
```
jgrassler: I'm not sure how stable that link is. Maybe it's better to copy the relevant parts to our documentation (we only need the time unit specification). The CMM operator's guide did that as well:

https://github.com/FujitsuEnablingSoftwareTechnologyGmbH/cmm-documentation/blob/master/DocSources/cmmoperator/install/c-cmmoinstcmm.xml#L118

xml/depl_nodes.xml
```diff
@@ -5002,6 +5008,461 @@ getent group postgres | cut -d: -f3</screen>
...
+        <para>
+	 Enable or disable email alarm notifications.
+        </para>
+      </listitem>
+   </varlistentry>
+  <varlistentry>
+      <term>smtp_host</term>
+      <listitem>
+        <para>
+	 SMTP smarthost.
```
jgrassler: Maybe add what it is being used for here (alarm notifications)

xml/depl_nodes.xml
dmpop: you mean *a problem with multiple plug-ins results*? In this case, I'd prefer to rewrite this as *a problem with the results from multiple plug-ins*. Does this make sense?

jgrassler: It does make sense, yes :-)

So sure, do it that way.

PR #75, opened 2017-05-11
Change 'my_mode" to one of three allowed modes

cjschroder:
bug https://bugzilla.suse.com/show_bug.cgi?id=1032137


taroth21:
Thanks! A docupdate entry for this bug is to arrive soon, right?

cjschroder:
Yes!

PR #73, opened 2017-05-10
new section documenting network-json-resolve

cjschroder:
bugzilla https://bugzilla.suse.com/show_bug.cgi?id=1032689

PR #76, opened 2017-05-11
Replace "teaming" with "team"

cjschroder:
https://bugzilla.suse.com/show_bug.cgi?id=1038741

I'll create docupdates entries for this, https://github.com/SUSE-Cloud/doc-cloud/pull/75, and https://github.com/SUSE-Cloud/doc-cloud/pull/73 after they are merged

### DIFF
--- a/xml/depl_conf_admin_crowbar.xml
+++ b/xml/depl_conf_admin_crowbar.xml
@@ -292,7 +292,7 @@
    linkend="sec.depl.req.network.modes"/> for details.
   </para>
   <para>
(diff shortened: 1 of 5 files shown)


taroth21:
Thanks a lot, Carla! For the records, some more info by cjschroder why we should discriminate between "teaming" and "team": 

The network modes are called 'single', 'dual', and 'team'. So I changed "teaming" to "team" where "teaming" was used as a network mode. 

However, "teaming" is a bonding mode in network.json.

PR #84, opened 2017-05-25
document RabbitMQ SSL configuration

cjschroder:
need screenshot of Crowbar GUI, some clarifications

@cmurphy


cjschroder:
https://bugzilla.suse.com/show_bug.cgi?id=1033135

cjschroder:
almost there--- allrightythen, I need extra help with this, O gurus.

- What is 'Generate (self-signed) certificates (implies insecure)' for? Does this actually generate new certs? When set to True is 'SSL Certificate is insecure' required to be True?
- What is 'SSL Certificate is insecure (for instance, self-signed)' for? Usually we advise that self-signed certs used exclusively in-house, with no external access, are safe to trust. When would this be set to "true"?
- What is 'SSL client CA file (used to validate rabbitmq server certificate)' for?

I know that is a lot of questions, but making this correct and thorough will save hardworking devs from a lot of pestering for answers. 

@cmurphy

cjschroder:
Ok, in other barclamps ''Generate (self-signed) certificates (implies insecure)' is only for testing, and it does generate certs. Is that the same here?

cmurphy:
> Ok, in other barclamps ''Generate (self-signed) certificates (implies insecure)' is only for testing, and it does generate certs. Is that the same here?

Yes, that option is the same as for other barclamps.

> What is 'SSL Certificate is insecure (for instance, self-signed)' for? Usually we advise that self-signed certs used exclusively in-house, with no external access, are safe to trust. When would this be set to "true"?

That is also the same as for other barclamps. For instance the keystone barclamps says "Set this option to true when using self-signed certificates to disable certificate checks. This setting is for testing purposes only and should never be used in production environments!"

> What is 'SSL client CA file (used to validate rabbitmq server certificate)' for?

This one is unique to the rabbitmq barclamp. This is the path where clients of rabbitmq will look for a CA bundle against which to validate the certificate presented by the rabbitmq server. This setting translates to the `[oslo_messaging_rabbit]/kombu_ssl_ca_certs` (renamed in more recent versions of openstack to `[oslo_messaging_rabbit]/ssl_ca_file`) configuration option that is present in every openstack config file. More detail on what it means upstream can be found here: https://docs.openstack.org/developer/oslo.messaging/opts.html#oslo_messaging_rabbit.ssl_ca_file

This parameter is required when insecure is false because the openstack services don't simply trust the system CA, so they must be told explicitly where to look for one.

To clarify, all of the other settings apply to how rabbitmq itself is configured, but this last one is applied to all of rabbitmq's clients.

cjschroder:
\o/ Many thanks, @cmurphy!

cjschroder:
replaced by https://github.com/SUSE-Cloud/doc-cloud/pull/94

PR #92, opened 2017-06-21
bsc#1042225 fix (hardware reqs for monasca server node)

dmpop:
@taroth21 @jgrassler Please take a look at this, and let me know if you have any feedback. Thank you!

### DIFF
--- a/xml/depl_require.xml
+++ b/xml/depl_require.xml
@@ -1408,11 +1408,11 @@
   </sect2>
 
 <sect2 xml:id="sec.depl.req.hardware.monnode">
-   <title>&contrnode;</title>
+   <title>&monnode;</title>
    <para>
      The Monasca agent roles, such as <literal>monasca-agent</literal> and
-     <literal>monasca-log-agent</literal> should be assigned to all nodes,
-     regardless of the nodes' other roles. However, the
+     <literal>monasca-log-agent</literal>, should be assigned to all nodes,
+     regardless of their other roles. However, the
      <literal>monasca-server</literal> role must be deployed on a dedicated physical machine
      referred as &monnode;. Hardware required for the &monnode; are as follows:
    </para>
@@ -1424,22 +1424,22 @@
      </listitem>
     <listitem>
      <para>
-       RAM: At least 32 GB, 64 GB or more recommended
+       RAM: At least 32 GB, 64 GB or more is recommended
      </para>
      </listitem>
     <listitem>
      <para>
-       CPU: At least 8 cores, 16 cores or more recommended
+       CPU: At least 8 cores, 16 cores or more is recommended
      </para>
      </listitem>
     <listitem>
      <para>
-       Hard Disk: SSD strongly recommended
+       Hard Disk: SSD is strongly recommended
      </para>
      </listitem>
    </itemizedlist>
    <para>
-     The required disk space is calculated using the following formula:
+     The following formula can be used to calculate the required disk space:
    </para>
    <screen>200 GB + ["number of nodes * "retention period" * ("space for log
    data/day" + "space for metrics data/day") ]</screen>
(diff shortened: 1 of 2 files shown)


xml/depl_require.xml
```diff
@@ -1406,6 +1406,110 @@
     For storage requirements, see <xref linkend="sec.depl.req.storage.hardware.store"/>.
    </para>
   </sect2>
+
+<sect2 xml:id="sec.depl.req.hardware.monnode">
+   <title>&monnode;</title>
+   <para>
+     The Monasca agent roles, such as <literal>monasca-agent</literal> and
```
taroth21: I think the information where to deploy which roles rather belongs to the Monasca barclamp section. The only piece of information I would mention here is probably the following: 
The &monnode; is a dedicated physical machine that runs the ``monasca-server``` role. It is used for &productname; Monitoring.  

In fact, even this definition (what the Monasca node is and what is it good for), should be covered in chapter 1 already (I will add it there as part of bsc#1042220 as soon as I have received some input by Johannes).

xml/depl_require.xml
```diff
@@ -1406,6 +1406,110 @@
...
+     <literal>monasca-log-agent</literal>, should be assigned to all nodes,
+     regardless of their other roles. However, the
+     <literal>monasca-server</literal> role must be deployed on a dedicated physical machine
+     referred as &monnode;. Hardware required for the &monnode; are as follows:
```
taroth21: Hardware required for -> The hardware requirements for [...]? (but I'm not sure about this).

xml/depl_require.xml
```diff
@@ -1406,6 +1406,110 @@
...
+     </para>
+     </listitem>
+   <listitem>
+     <para>
+      20 bytes in size
+     </para>
+     </listitem>
+   </itemizedlist>
+   <para>
+     Please note that the formula provides only a rough estimation of the
```
taroth21: I would put this information directly after the formula and remove the 'Please note':
The formula provides only...

xml/depl_require.xml
jgrassler: Yes, either "The hardware requirements for [...] are as follows" or "The following Hardware is required for [...]. Right now this is a mix of both.

xml/depl_require.xml
```diff
@@ -1406,6 +1406,110 @@
...
+     required disk space. Depending on the number of monitored nodes, other
```
jgrassler: I'd reformulate this a bit and make it more concrete: "The mix of  services which actually run on your OpenStack nodes, your cloud's usage pattern and whether any or all services have debug logging enabled may lead to higher or lower figures than what the formula yields."

xml/depl_require.xml
dmpop: Makes sense. Fixed.

xml/depl_require.xml
dmpop: Oops, my bad! Fixed.

xml/depl_require.xml
dmpop: Sounds sensible, but I think the assumptions and recommended values take precedence over the note. In other words, it's more important to clarify the formula's values and basis before providing additional info. Thoughts?

xml/depl_require.xml
dmpop: How about this, then?

_There are several factors that can affect disk space requirements. This
includes the exact combination of services that run on your OpenStack nodes,
actual cloud usage pattern, and whether any or all services have debug logging enabled._

jgrassler: Yes, that looks good. Thanks!

xml/depl_require.xml
jgrassler: Are you sure? The diff still shows the old version ("Hardware required for the &monnode; are as follows").

xml/depl_require.xml
taroth21: Fine with me, too. :)

xml/depl_require.xml
taroth21: Fine with me, too. :)

PR #95, opened 2017-07-03
Update 'NFA Exports' in cinder barclamp

cjschroder:
https://bugzilla.suse.com/show_bug.cgi?id=1043385


cjschroder:
Good enough, merging!

PR #111, opened 2017-07-31
   Document nova-compute-ironic for Ironic barclamp

cjschroder:
https://bugzilla.suse.com/show_bug.cgi?id=1050587


cjschroder:
Thank you!

PR #118, opened 2017-08-14
ironic: Update driver information

cmurphy:
The pxe_ssh driver was deprecated in the Newton release of OpenStack
ironic and it is likely to be removed in the Pike release. This patch
removes reference to it as a recommended driver, leaving only the
pxe_ipmitool driver as recommended. Additionally, change the upstream
driver reference to point to the official documentation rather than the
wiki, which not rigorously maintained and is outdated.


cjschroder:
👍 
Thanks Colleen.

PR #122, opened 2017-08-21
Update Glance option is-public/visibility

cjschroder:
since glance api 2.0, --is-public is deprecated, replaced by --visibility

http://doccomments.provo.novell.com/comments/30137


cjschroder:
Thanks @nicolasbock

PR #123, opened 2017-08-21
Correct glance option --is-visible/--visibility

cjschroder:
--visibility replaced --is-visible in glance api 2.0

http://doccomments.provo.novell.com/comments/30137


cjschroder:
nova flavor-create still uses  --is-visible

PR #124, opened 2017-08-21
Correct glance option --is-visible/--visibility

cjschroder:
--visibility replaced --is-visible in glance api 2.0


cjschroder:
That should be the last one!

cjschroder:
Thanks @nicolasbock.

PR #127, opened 2018-01-30
Ex-Helion Installation/Planning content 

sknorr:
This branch tries to omit all changes that are not related to the converted version of the docs in question.
This means:
* there should be no trace of the DITA files in here
* there should be no large videos or spreadsheets in here

There should also be as little  drama as possible in here, i.e. the goal was to not copy all slip-ups and mistakes made on the way here.


xml/installation-kvm_xpointer.xml
```diff
@@ -530,7 +530,7 @@ git commit -m "My config or other commit message"</screen>
   </itemizedlist>
   <para>
    If you chose this method for installing your baremetal hardware, skip
-   forward to the <xref linkend="install_kvm"/> step.
+   forward to the <!-- FIXME: this seems to be the wrong reference <xref linkend="install_kvm"/> - sknorr, 2018-01-19 --> step.
```
dmpop: There is a handful of FIXMEs in this file that need to be, er, fixed. I created a Trello card for this: https://trello.com/c/fYtlCjok

PR #128, opened 2018-02-04
Feature/mysql mariadb

csymons-suse:
changed MySQL to MariaDB & mysql to mariadb; changed orderedlist to procedure where appropriate


sknorr:
Carl, we can't take this branch the way it is right now. The issue is that you started this feature branch from the hos-docs-rename branch and not from develop and hence it would bring a lot of changes that I'd rather like to avoid in `develop`. However, this is not hard to fix:

1. locally on the CLI, do `git check out develop`
2. open another new feature branch: `git checkout -b feature/mysql-mariadb2`
3. copy the two commits from this branch that we want (69a9396, 0234247)  to your new branch: `git cherry-pick 69a9396 0234247`
4. then push and create a new PR.

Step 3 can be tricky because in some cases git can not cherry-pick successfully. If it does not work, just ask me for help.

PR #129, opened 2018-02-05
use entities for companies/products (further to bsc#1079065)

sknorr:
this was initially only supposed to fix the remaining parts of bsc#1079065 but quickly got out of hand. @dmpop could you look a have?


sknorr:
The commits here were a bit of a mess, but I merged them now anyway since:
a) stuff builds now
b) reworking them all would be a huge mess as well.

PR #161, opened 2018-02-23
rewrite GUI instructions

csymons-suse:
still several open issues that have been referred to Subject Matter Experts
interim update so that GUI installer team has something to look at and work with

### DIFF
--- a/xml/installation-installation-gui_installer.xml
+++ b/xml/installation-installation-gui_installer.xml
@@ -321,7 +321,7 @@
  </para>
  </section>
  <section>
-<title>Run the Install UI to nstall our cloud</title>
+<title>Run the Install UI to install your cloud</title>
  <para>
   The Install UI is found on the &lcm; node at:
   <uri>http://<replaceable>LIFECYCLE_MANAGER_IP</replaceable>:3000</uri>
@@ -343,6 +343,28 @@
  <para>
 	 The first page of the Install UI shows the general installation process and a reminder to gather some information before beginning. Clicking the <guimenu>Next</guimenu> button brings up the Model Selection page.
  </para>
+ <!-- FIXME: add proper PNGs
+ <informalfigure vendor="hpe">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_intro_oem_hpe.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_intro_oem_hpe.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+ <informalfigure vendor="suse">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_intro.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_intro.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+ -->
  <para>
 	 <emphasis role="bold">Choose an OpenStack Cloud Model</emphasis>
  </para>
@@ -352,6 +374,48 @@
  <para>
 	 Selecting a model will determine the base template from which the cloud will be deployed. Models can be adjusted later in the process, though selecting the closest match to your requirements reduces the effort required to deploy your cloud.
  </para>
+ <!-- FIXME: add proper PNGs
+ <informalfigure vendor="hpe">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_select_model_oem_hpe.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_select_model_oem_hpe.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+ <informalfigure vendor="suse">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_select_model.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_select_model.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+	<informalfigure vendor="hpe">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_model_details_oem_hpe.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_model_details_oem_hpe.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+ <informalfigure vendor="suse">
+    <mediaobject>
+      <imageobject role="fo">
+        <imagedata fileref="installer_ui_model_details.png" width="75%" format="PNG"/>
+      </imageobject>
+      <imageobject role="html">
+        <imagedata fileref="installer_ui_model_details.png"/>
+      </imageobject>
+    </mediaobject>
+  </informalfigure>
+ -->
  <para>
 	 <emphasis role="bold">Cloud Model to Deploy</emphasis>
  </para>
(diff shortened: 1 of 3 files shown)


xml/installation-installation-gui_installer.xml
```diff
@@ -8,51 +8,72 @@
...
-  template) and edit the contents to reflect your environment.
+  When installing via the GUI, you begin with one of 8 input model examples
+  (essentially cloud configuration templates) and are guided step-by-step to
+  edit the contents to reflect your environment.
  </para>
  <para>
-  However, the full range of customizations is not possible via the GUI as some
-  configuration items are not exposed. Below is the list of what can and cannot
-  be changed:
+  The full range of customizations is not possible using the GUI. Some customizations require editing the yml files directly, which can be done in the "review configuration files" section of the GUI. Below is the list of what can be changed via GUI widgets, followed by a list of customizations that require direct editing of the configuration files to change.
```
jeremy-moffitt: not sure if you need line breaks here or if the long line is ok... guess that depends on the doctool

xml/installation-installation-gui_installer.xml
```diff
@@ -91,16 +112,10 @@
   </listitem>
  </itemizedlist>
  <para>
-  <emphasis role="bold">Changes to the following items cannot be made:
-  </emphasis>
+  <emphasis role="bold">Changes to the following items can only be made by manually editing the associated yml files on the 'Review Configuration' page of the GUI:</emphasis>
```
jeremy-moffitt: same question as above about line breaks, unclear if long lines are ok

xml/installation-installation-gui_installer.xml
```diff
@@ -340,26 +351,19 @@
...
   The username and password should be what was set in
   <xref linkend="sec.kvm.setup_deployer"/>. For example:
  </para>
-<screen>ssh -v -N -L 8080:&lt;lifecycle manager_IP&gt;:79 &lt;Username&gt;@&lt;lifecycle manager_IP&gt;</screen>
+<screen>ssh -v -N -L 8080:&lt;lifecycle manager_IP&gt;:3000 &lt;Username&gt;@&lt;lifecycle manager_IP&gt;</screen>
  <para>
-  Then on your local machine, the one you are tunneling from, point your web
-  browser to: <uri>http://localhost:8080/dayzero</uri>
+  Then on your local machine (the one you are tunneling from) point your web
+  browser to: <uri>http://localhost:8081</uri>
```
jeremy-moffitt: this should be 8080 if the ssh tunnel command above is used... its not the port number of the installer, but rather of the ssh tunnel... so the "ssh -v -N -L 8080" command above means that this url should still use "8080" as the port rather than "8081"

xml/installation-installation-gui_installer.xml
```diff
@@ -315,92 +333,184 @@
...
+<para>
+Choose servers on which &hlinux; will be installed
+If an OS has not previously been installed on the servers that make up the cloud
+configuration, the OS installation page allows for Cobbler to deploy &hlinux;
+on servers in the cloud configuration. Enter credentials, select servers and 
+click <guimenu>Install</guimenu> to deploy &hlinux; to these servers. An 
+installation log and progress indicators will be displayed.
+</para>
+<para>
+When the installation is complete, a Server and Server Role Summary is
```
jeremy-moffitt: Maybe "OS installation" instead of "installation" here

xml/installation-installation-gui_installer.xml
```diff
@@ -315,92 +333,184 @@
...
+displayed. Any missing settings are identified. There is also an opportunity to
+Manage Cloud Settings without having to edit YAML files directly. Changes at
+this point may cause the configuration to fail validation.
+</para>
+<para>
+Server and Role Summary:
+When the installation is complete, a Server and Server Role Summary is
+displayed. It shows which servers have been assigned to each role, and provides
```
jeremy-moffitt: the first few lines of this block of text is duplicated just above

xml/installation-installation-gui_installer.xml
```diff
@@ -410,32 +786,21 @@
...
-    Visit the <xref linkend="ceph_overview"/> for Ceph post-install steps.
-   </para>
-  </listitem>
- </itemizedlist>
+<title>Use the Install UI securely</title>
+ <para>
+  The Install UI is a web-based application that runs on the &lcm; node and is
+  implemented in ReactJS and Python. This application runs behind Apache 2.0 on
+  port 3000. Browser refresh is not necessary if you are using the browser
+  connected directly to the Install UI.
```
sknorr: ~Nobody cares about these implementation details -- ReactJS, Python, and Apache don't need to be mentioned.  (The people that do care will just take a look at the source code.)

I also don't understand the bit about refreshing the browser/page(?).

xml/installation-installation-gui_installer.xml
```diff
@@ -410,32 +786,21 @@
...
+ </para>
+ <para>
+  You should access the Install UI through Apache. An example URL looks like
+  this: http://&lt;lifecyclemanager-host-ip&gt;:3000.
```
sknorr: Put the URL in a `<literal>` and use `<replaceable>`: `<literal>http://<replaceable>LIFE_CYCLE_MANAGER_HOST_IP</replaceable>:3000</literal>`

Also, is that port number correct? If I understand Jeremy correctly, it should be 8080 instead..?

jeremy-moffitt: this is complicated... if the user is accessing the installation UI directly on the server that its running on, then they will use port 3000 ... if they're using an ssh tunnel to map port 8080 on their localhost to port 3000 on the lifecycle manager, then they'll need to use port 8080... I believe in this context that port 3000 is correct, but this does make me think that we need to clarify this a bit

xml/installation-installation-gui_installer.xml
jeremy-moffitt: Because of the network isolation of the deployer from the rest of a customers cloud, they may access the Installation UI 2 ways:
1. by logging into the lifecycle manager host directly and launching a browser from there
2. by creating an ssh tunnel from their local server to the lifecycle manager host

during the deployment process the network on the lifecycle manager will be restarted. If the customer is connected via method 2 above , then they will lose that connection when the network restarts, and will need to refresh/reload their browser after re-establishing their ssh tunnel. If they launched the browser directly from the lifecycle manager, then they will not have to manually refresh .

xml/installation-installation-gui_installer.xml
sknorr: Ah, ok -- since I only looked at the sections that changed, there is a good chance that I missed that. Thanks for the explainer!

xml/installation-installation-gui_installer.xml
```diff
@@ -315,92 +300,483 @@
...
+    </mediaobject>
+  </informalfigure>
+-->
+ <para>
+	 <emphasis role="bold">Review Configuration Files</emphasis>
+ </para>
+ <para>
+	 Advanced editing of the cloud configuration can be done on the "Review Configuration Files" page. Individual yml and J2 files can be edited directly with the embedded editor in the Model and Templates and Services tabs. The Deployment tab covers Wipe Data Disks, Encryption Key and Verbosity Level.
+ </para>
+ <!-- FIXME: add proper PNGs
```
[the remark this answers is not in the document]
jeremy-moffitt: The page in question has a 3 tabs, 2 of which have lists of files that may be edited, and the 3rd has an embedded form... Without seeing this built with the files embedded, I can only guess that this is following the sequence of showing each tab and the ability to edit those files... we probably could intersperse some text between them for additional context

csymons-suse:
the image file names are the ones that Jeremy used. If those are pulled into images/src/png, then the document comment tags can just be removed.

csymons-suse:
git cherry-pick c8cf0d2c doesn't work
fatal: bad revision 'c8cf0d2c'
I have all the files, but I don't see a way to get them into this commit without risking an error. I considered just adding them to images/src/png, but it seems risky.
Please advise.

jeremy-moffitt:
is there a reason we can't just merge the changeset with the images and then rebase this?

xml/installation-installation-gui_installer.xml
```diff
@@ -6,53 +6,46 @@
...
- <para>
-  When installing via the GUI, you begin with a model (or a cloud configuration
-  template) and edit the contents to reflect your environment.
- </para>
- <para>
-  However, the full range of customizations is not possible via the GUI as some
-  configuration items are not exposed. Below is the list of what can and cannot
-  be changed:
- </para>
+ <title>Installing with the Install UI</title>
```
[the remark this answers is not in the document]
csymons-suse: We had a go-around with the Cloud Day0 install squad. Install UI won.

xml/installation-installation-gui_installer.xml
```diff
@@ -129,7 +118,7 @@
 <section>
 <title>Before you begin</title>
  <para>
-  Before you run the GUI installer to install your cloud, there are a number of
+  Before you run the Install UI to install your cloud, there are a number of
   things you need to do:
  </para>
  <orderedlist>
```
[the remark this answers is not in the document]
csymons-suse: It's just a list; doesn't have to be done in a particular order.

sknorr:
@csymons-suse
> git cherry-pick c8cf0d2 doesn't work fatal: bad revision 'c8cf0d2c' [...] Please advise.

Ah, my bad -- I overlooked that Jeremy was working from a branch in his own cloned repo, hence the commit is not available in this central repo. Anyway, I followed Jeremy's idea...


@jeremy-moffitt 
> is there a reason we can't just merge the pull request with the images and then rebase this? I just rebased that PR to get it up to date with the latest develop branch

Yeah, you are right -- I did just that.

PR #174, opened 2018-02-27
Feature/remove no longer needed

csymons-suse:
removing various references to vsa and other content no longer needed


sknorr:
Thanks!

PR #182, opened 2018-03-05
Fix table xml:id="services_components"

tomschr:
This PR contains the following changes:

* Make sure all the `morerows` attributes are correct

### DIFF
--- a/xml/planning-architecture-input_model-other_topics-services_components.xml
+++ b/xml/planning-architecture-input_model-other_topics-services_components.xml
@@ -21,7 +21,7 @@
    </thead>
    <tbody>
     <row>
-     <entry morerows="2">Compute</entry>
+     <entry morerows="2" valign="middle">Compute</entry>
      <entry>Virtual Machine Provisioning</entry>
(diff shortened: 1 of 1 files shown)


sknorr:
This changes the table in the wrong way -- it adds empty cells in the wrong position. See #183...

xml/planning-architecture-input_model-other_topics-services_components.xml
```diff
@@ -129,6 +135,13 @@ swift-ring-builder
 swift-rsync</screen>
      </entry>
     </row>
+    <row>
+     <entry>HPE Virtual Storage Array</entry>
+<!--     <entry></entry>-->
+     <entry>vsa-storage</entry>
+     <entry><screen>cmc-service
+vsa</screen></entry>
+    </row>
```
dmpop: *HPE Virtual Storage Array* entry actually shouldn't be in the table. Could you please remove it? Thanks! :smile:

tomschr:
The table should now be correct.

PR #184, opened 2018-03-05
Feature/ui install fixups

csymons-suse:
edit Install UI section based on 
bsc#1083342 - [doc] 7.1 Before You Begin has wrong information

### DIFF
--- a/xml/installation-installation-gui_installer.xml
+++ b/xml/installation-installation-gui_installer.xml
@@ -170,11 +170,14 @@ sudo sed -i -e "s,.*solver.allowVendorChange.*,solver.allowVendorChange = true,"
      </step>
      <step>
       <para>
-       Mount the repo as <literal>DC8S</literal>:
+       Add the repository as <literal>cloud-install-repo</literal>:
       </para>
-<!-- FIXME: This probably needs to be profiled for HPE since the HPE
-      ISO has a different name. - sknorr, 2018-02-28 -->
-<screen>sudo zypper ar -G "iso:///?iso=$(ls ~/stack/SUSE-OPENST*.iso)" DC8S</screen>
+<screen vendor="suse">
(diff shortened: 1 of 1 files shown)


xml/installation-installation-gui_installer.xml
```diff
@@ -159,21 +159,20 @@
...
-  <screen>
-<?dbsuse-fo font-size="0.65em"?>
-sudo sed -i -e "s,.*solver.allowVendorChange.*,solver.allowVendorChange = true," /etc/zypp/zypp.conf
-  </screen>
-     </step>
-     <step>
-      <para>
-       Mount the repo as <literal>DC8S</literal>:
+      <para vendor="suse">
+       Mount the repository as <literal>&cloud;</literal>:
```
sknorr: You will not make friends with your users if you advise that they mount anything into a dir with a super-long name with spaces in the middle. I think "DC8S", despite being cryptic is a much better name, though, how about just "cloud-install-repo" or so? -- a bit long, but easy to easy to memorize and we won't ever need to update that name because there is no version number in there.

Then, you can also remove the vendor profiling from this para.

xml/installation-installation-gui_installer.xml
```diff
@@ -159,21 +159,20 @@
...
       </para>
-      <!-- FIXME: This probably needs to be profiled for HPE since the HPE
-      ISO has a different name. - sknorr, 2018-02-28 -->
-      <screen>sudo zypper ar -G "iso:///?iso=$(ls ~/stack/SUSE-OPENST*.iso)" DC8S</screen>
+      <screen vendor="suse">
```
sknorr: The following screens should not be indented. And you don't need to the font-size PI, these are less than 80 characters.

sknorr:
One issue I still see is that actually the word mount is used the wrong way -- the repo is not actually mounted there, it is added to the repository list under a certain alias.

Also, before you merge, please squash.

PR #197, opened 2018-03-14
Fix bsc#1085066

dmpop:
Removed an obsolete step


sknorr:
Ok, we decided to not wait, since we have to build packages again soon.

PR #210, opened 2018-03-24
HP, VSA references; images; entities; edits

csymons-suse:
remove HPE Helion from images
change HPE references to vendor entity references
remove helion and vcp references
change proxy from HP.tld to example.org
change HOS download site
change VSA to SES in images


xml/installation-installation-ironic-install_ironic_overview.xml
```diff
@@ -2,17 +2,16 @@
...
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <title>Installing Baremetal (Ironic)</title>
  <para>
-  Bare Metal as a Service is enabled in this release for deployment of nova
-  instances on bare metal nodes using flat networking. HPE DL and SL line of
-  servers are supported.
-  <!-- FIXME: HPE mention etc. -->
+  Bare Metal as a Service is enabled in this release for deployment of Nova
+  instances on bare metal nodes using flat networking.
+  <phrase vendor="hpe">&vendor;</phrase> DL and SL line of servers are
```
sknorr: This is a rather novel approach to profiling, as the output is the following:

* HPE: `[...] networking. HPE DL and SL line [...]`
* SUSE: `[...] networking. DL and SL line [...] `

I other words, yes, you removed the HPE mention in the SUSE documentation but it does not make any sense anymore. Depending on the circumstances (i.e. what the product team is saying), you could do one of the following:

* remove the whole sentence in the SUSE version (because in general we don't necessarily push our customers to buy HPE hardware over say Dell or Lenovo).
* just leave the sentence as is, but maybe use the `&hpe;` entity (without any profiling!!)

Note that you have done added this weird version of `&vendor;` in a lot of places in this PR and I have not added a comment for each occurrence.

xml/installation-installation-ironic-ironic_multi_control_plane.xml
```diff
@@ -23,16 +22,16 @@
   </para>
   <para>
    You need to set the environment variable
-   <emphasis role="bold">OS_REGION_NAME</emphasis> to the ironic region in
-   baremetal control plane. This will setup the ironic flat networking in
-   neutron.
+   <emphasis role="bold">OS_REGION_NAME</emphasis> to the Ironic region in
```
sknorr: If you are replacing such things -- please add an entity for "Ironic" and us it, also use the existing entity for "Neutron" (and those for "Swift", "Glance", etc. later on).

xml/installation-networking-l2gw5930.xml
```diff
@@ -197,18 +197,18 @@ ip address 10.1.1.10 255.255.255.0</screen>
 ip address 10.10.83.3 255.255.255.0</screen>
     <para>
      The 5930 switch has a fortygigE1/0/5 interface to which a splitter cable
-     is connected that splits the network into four tengigEthernet
+     is connected that splits the network into four tengig Ethernet
```
sknorr: I don't think splitting that word makes sense, given that it used as the interface name here, apparently. Rather put it in a literal element.

PR #212, opened 2018-03-26
Feature/helion lingers

csymons-suse:
Resolve bsc#1083228 - Screenshots still refer to VSA


csymons-suse:
I agree. The imperative was "remove all mentions of certain constructs". 
At the time, there were no provisions for doing realistic screenshots.

PR #219, opened 2018-04-08
clean up security/Barbican; fold in new release features file

csymons-suse:
remove new release section; add entity references

combined 2 previous commits


xml/security-barbican.xml
```diff
@@ -163,32 +130,32 @@
...
-     No changes are needed for Barbican to be enabled. When installing your
-     cloud, you should use the input model definitions which already define the
-     necessary Barbican components. When using the pre-defined input model
-     files that come with &kw-hos-phrase;, nothing else needs to be done in
-     those files.
+     For new installations, no changes are needed for &secret_store; to be
+     enabled. When installing your cloud, you should use the input models which
+     already define the necessary &secret_store; components. When using the
+     pre-defined input model files that come with &kw-hos-phrase;, nothing else
+     needs to be done in those files.
```
sknorr: Idea for condensing this paragraph into two easy-to-read sentences:

```
During the installation, choose an input model that defines the &secret_store;
components. There are no further changes necessary to enable &secret_store;.
```

xml/security-barbican.xml
```diff
@@ -197,35 +164,43 @@
...
- <section>
-  <title>Error Received if Root Logical Volume is Too Small</title>
+ <section xml:id="sec.barbican.kmip">
+  <title>Master Key Configuration</title>
   <para>
-   Barbican currently supports databases and KMIP as its secret store back-ends
-   In &ostack-bare; upstream additional back-ends are available, such as the
-   PKCS11 and dogtag plug-ins, but they are not tested or supported by
+   &secret_store; currently supports databases and KMIP as its secret store
+   back-ends In &ostack-bare; upstream additional back-ends are available, such
```
sknorr: add a full stop after `back-ends.`, add a comma between `upstream, additional` (this commit is not at fault here, it just makes the issues more obvious.)

xml/security-barbican.xml
```diff
@@ -484,22 +545,22 @@ ansible-playbook -i hosts/localhost ready-deployment.yml</screen>
       </entry>
       <entry>
        <para>
-        Barbican specific role with privilege to create/modify/list keys,
-        certificates.
+        &secret_store; specific role with privilege to
+        create/modify/list/delete keys, certificates.
```
sknorr: please replace by `create, modify, list, and delete` -- the current version looks weird and will lead to odd line breaks. (again, not your fault)

xml/security-barbican.xml
```diff
@@ -197,35 +164,43 @@
...
-   master key.
+   the initial &secret_store; configuration, you must generate and configure
+   this master key.
+  </para>
+  <para>
+   When &secret_store; is used with <literal>simple_crypto_plugin</literal> as
+   its secret store back-end, its master key needs to be defined
+   <emphasis role="bold">before initial deployment</emphasis>. If no key is
+   specified before deployment, the default master key is
+   used&mdash;<emphasis role="bold">this practice is discouraged</emphasis>.
```
sknorr: As style thing: we generally avoid using `<emphasis role="bold">`. Instead, I would recommend creating a note here, and speaking a bit more directly to the user:

```
<important>
  <title>Undefined Master Key</title>
  <para>
    If you do not define a master key before deployment, the default master key is
    used. We recommend against this practice.
  </para>
</important
```

Ideally, we could write something about the consequences, i.e. if this is a security issue or so, but at least I don't know details here.

xml/security-barbican.xml
```diff
@@ -271,19 +246,101 @@ ansible-playbook -i hosts/localhost ready-deployment.yml</screen>
     </para>
    </step>
   </procedure>
- </section>
- <section xml:id="sec.barbican.kmip">
-  <title>KMIP Plug-in Support</title>
+  <bridgehead renderas="sect4">Upgrade Master Key Configuration</bridgehead>
```
sknorr: Please avoid adding more bridgeheads to these documents -- use real sections instead. If there is a nesting level issue, bridgeheads don't really avoid it, they just hide it. We will need to reorganize large parts of these documents at some point anyway, because of their completely impractical nesting level.
Also, for titles, use gerund (`ing`).

xml/security-barbican.xml
```diff
@@ -271,19 +246,101 @@ ansible-playbook -i hosts/localhost ready-deployment.yml</screen>
...
+  <procedure>
+   <step>
+    <para>
+     Check the master key.
+    </para>
+    <para>
+     If a master key is already defined, check
```
sknorr: Be more direct if this a user task: `If you have not defined a master key, ...`

xml/security-barbican.xml
```diff
@@ -271,19 +246,101 @@ ansible-playbook -i hosts/localhost ready-deployment.yml</screen>
...
+     <filename>~/openstack/ardana/ansible/roles/barbican-common/vars/barbican_deploy_config.yml</filename>
+     for <literal>barbican_customer_master_key</literal> value. If the value
```
sknorr: better: `for the value of <literal>barbican_customer_master_key</literal>. If the value`

sknorr:
@csymons-suse Hm, did you really mean to remove the commit from here and close this? (I just made a backup branch locally, in case that was not your intention.)

PR #226, opened 2018-04-12
SP2 to SP3

dmpop:
Replaced SP2 with SP3 where necessary, as per https://bugzilla.suse.com/show_bug.cgi?id=1089145


kbaikov:
Are you sure those are all the mentions of SP2?
If i grep the local doc-cloud repo i find much more than 5 files.
But perhaps i am on a different branch/version.

dmpop:
I left out a handful of SP2 mentions related to upgrading to SP3. Otherwise it should be OK.

dirkmueller:
@sknorr The only ironic image that we support right now is the one we provide as part of the product. for the supported KVM Guest OS'es, that information should be removed and point to the supported guest OS'es from SLES - we're inheriting the KVM packages from there and information shouldn't be duplicated.

sknorr:
@dmpop Please don't make additional commits à la "Integrated additional feedback"/"Fixed typos" etc, just squash them into the original commit. Also, please make sure to include the bsc# in the commit message, not just in the PR description. (Both fixed now.)

PR #228, opened 2018-04-12
replace GUI Installer screenshots

csymons-suse:
remove vendor profiling
as directed by Day0 squad (Gloria Gu; Jeremy Moffitt)
commit cherrypicked from
2f1ce644067fee225a78d27c84e5c1d2b30218ff


sknorr:
This won't be necessary any more -- we had this commit in another branch already. -> Closed.

PR #238, opened 2018-04-20
migration: Add important note to use latest doc (bsc#1090232)

toabctl:
Given that the migration might be only supported to a GM (without
updates installed), mention that the documentation should not be taken
from GM (but the online version which is GM+Updates).


rsalevsky:
What about HPE customers? I know that HPE has a own Doc page but I'm not sure where it is located.

sknorr:
Thank you!

PR #247, opened 2018-04-26
AppArmor section removed

csymons-suse:
requested by Bryan Stephenson; email 2018-04-25
works for Debian, but not yet done for SLES


sknorr:
I don't know if this is a good idea -- AppArmor is a technology that we usually push at SUSE. Could you check back with Brain whether the documentation from SLES (see https://susedoc.github.io/doc-sle/SLE12SP3/SLES-security/html/part.apparmor.html) applies? If so, we could link there.

PR #248, opened 2018-04-26
Changed more instances of private_data_metadata.yml to private_data_metadata_ccp.yml

bryanstephenson:
Changed more instances of private_data_metadata.yml to private_data_metadata_ccp.yml in four files. The old named file, private_data_metadata.yml, no longer exists so we need to update all references to it to the new name, private_data_metadata_ccp.yml.  Because I was not sure it would have no side effects I did not change the XML ID :  

 "xml:id="private_data_metadata_yml"

so this ID does not include the "_ccp" suffix.


xml/security-horizon_ssl_auth.xml
```diff
@@ -212,7 +212,7 @@
      The following is an example of a properly formatted DN for a certificate
      issuer named "intermedia".
     </para>
-<screen>/C=US/ST=California/O=HPE/OU=Engineering/CN=intermediateca/emailAddress=intermediateca@hpe.com</screen>
+<screen>/C=US/ST=California/O=SUSE/OU=Engineering/CN=intermediateca/emailAddress=intermediateca@hpe.com</screen>
```
sknorr: Don't know about this one -- we usually use example companies instead of "SUSE". Also there is the discrepancy between "intermedia"/"intermediateca" (Spanish??) and the mail address is still \@hpe.com.
Ordinarily, at least for the mail we'd use something like maybe "certificates\@example.org" and the company name could be "Example, LLC" perhaps. (Please adapt my suggestions where they don't make sense.)

If you really want to use a real organization, you could also use `&vendor;` which will switch between SUSE and HPE depending on document variant.

sknorr:
Thanks for the update!

However, there were still some build issues that I overlooked initially. I have opened https://github.com/SUSE-Cloud/doc-cloud/pull/252 with your changes + the build fixes in it (fixes a cross-reference and a file name reference in your "instances of private_metadata" commit).

If you're fine with that -- I would merge that version instead.

fsundermeyer:
Superseded by PR  #252 which has been merged. Closing.

PR #260, opened 2018-05-02
Update master to released SOC 7 documenation

fsundermeyer:


fsundermeyer:
Oops, last commit fails test

PR #266, opened 2018-05-09
More notes/warnings

sknorr:
Adds notes that were skipped in the initial conversion due to a Dietrich bug.


xml/bura-freezer_agent.xml
```diff
@@ -78,6 +85,12 @@
       </para>
      </listitem>
     </itemizedlist>
+    <note>
+     <para>
+      It is recommended that you use snapshots if the mode is MySQL, SQL
+      Server, or Mongo.
```
sknorr: Is MongoDB still supported? (I suspect I need to do MySQL->MariaDB here anyway.)

PR #278, opened 2018-05-22
Remove CPU models

csymons-suse:
We support all CPU models; no need to spell out specifically
corrected as requested by TR Bosworth, 
with input from Mike Latimer (SES Engineer) and 
Michal Svec (Product Manager)

### DIFF
--- a/xml/operations-compute-using_flavor_metadata.xml
+++ b/xml/operations-compute-using_flavor_metadata.xml
@@ -8,154 +8,45 @@
 <section xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="topic_vhs_12v_vw">
  <title>Using Flavor Metadata to Specify CPU Model</title>
  <para>
-  In &kw-hos; Nova, the <literal>ComputeCapabilitiesFilter</literal>
-  scheduling filter matches the exact CPU model of the compute host and the
-  requested compute instance model. It doesn't entertain the libvirt's
-  capability of emulating the hosts CPU model to its guests if the guest's CPU
-  model is older or equal to that of the hosts.
+  <literal>Libvirt</literal> is a collection of software used in &ostack; to
+  manage virtualization. It has the ability to emulate a host CPU model in a
+  guest VM. In &kw-hos; &o_comp;, the ComputeCapabilitiesFilter limits this
+  ability by checking the exact CPU model of the compute host against the
+  requested compute instance model. It will only pick compute hosts that have
+  the <literal>cpu_model</literal> requested by the instance model, and if the
+  selected compute host does not have that <literal>cpu_model</literal>, the
+  ComputeCapabilitiesFilter moves on to find another compute host that matches,
+  if possible. Selecting an unavailable vCPU model may cause &o_comp; to fail
+  with <literal>no valid host found</literal>.
  </para>
  <para>
-  To assist, we have introduced a new Nova scheduler filter which captures the
-  cpu_models as a subset of a particular CPU family. The filter determines if
-  the host CPU model is capable of emulating the guest CPU model by
-  maintaining the mapping of the vCPU models per vendor (currently Intel and
-  AMD) and compares it against the 'model' of the host.
+  To assist, there is a &o_comp; scheduler filter that captures
+  <literal>cpu_models</literal> as a subset of a particular CPU family. The
+  filter determines if the host CPU model is capable of emulating the guest
+  CPU model by maintaining the mapping of the vCPU models and comparing it with
+  the host CPU model.
  </para>
  <para>
-  There is a current limitation which is when a particular cpu_model is
-  requested through <literal>hw:cpu_model</literal> via a compute flavor, the
-  cpu_mode will be set to "custom". This mode makes it so that a persistent
-  guest virtual machine will see the same hardware no matter what host
-  physical machine the guest virtual machine is booted on. This allows easier
-  live migration of virtual machines. Because of this limitation, only some of
-  the features of a CPU are exposed to the guest. Requesting a particular CPU
-  feature is not currently supported.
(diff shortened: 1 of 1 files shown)


xml/operations-compute-using_flavor_metadata.xml
```diff
@@ -8,143 +8,29 @@
...
-  There is a current limitation which is when a particular cpu_model is
-  requested through <literal>hw:cpu_model</literal> via a compute flavor, the
-  cpu_mode will be set to "custom". This mode makes it so that a persistent
-  guest virtual machine will see the same hardware no matter what host
-  physical machine the guest virtual machine is booted on. This allows easier
-  live migration of virtual machines. Because of this limitation, only some of
-  the features of a CPU are exposed to the guest. Requesting a particular CPU
-  feature is not currently supported.
+  There is a limitation when a particular cpu_model is requested through
+  <literal>hw:cpu_model</literal> via a compute flavor, the cpu_mode will be
```
sknorr: Should the second occurrence at the end of the line really be "cpu_mode" or should it be "cpu_mode**l**" as well?

xml/operations-compute-using_flavor_metadata.xml
```diff
@@ -8,143 +8,29 @@
...
  <para>
-  In &kw-hos; Nova, the <literal>ComputeCapabilitiesFilter</literal>
+  In &kw-hos; &o_comp;, the <literal>ComputeCapabilitiesFilter</literal>
   scheduling filter matches the exact CPU model of the compute host and the
-  requested compute instance model. It doesn't entertain the libvirt's
-  capability of emulating the hosts CPU model to its guests if the guest's CPU
-  model is older or equal to that of the hosts.
+  requested compute instance model. It does not support the libvirt capability
+  of emulating the host CPU model to its guests if the guest's CPU model is
+  older or equal to that of the hosts.
```
sknorr: Mmh, this final sentence is weird and hard to understand. Does the following replacement make sense? `While libvirt supports emulating guest CPUs that are different from the host CPU, &o_comp; does not support that feature.`

xml/operations-compute-using_flavor_metadata.xml
csymons-suse: "mode" is better given other content there. 
"...the cpu_mode will be
  set to <literal>custom</literal>. This *mode* makes it so that a persistent
  guest virtual machine..." The latter sentence doesn't make sense if "model" is used.

xml/operations-compute-using_flavor_metadata.xml
csymons-suse: tweaked that section. We are gonna have work to do...there are quite a few sentences that are weird and hard to understand. I've been doing my best to ignore them in favor of fixing Very Wrong Stuff.

xml/operations-compute-using_flavor_metadata.xml
```diff
@@ -8,143 +8,29 @@
...
  <para>
-  In &kw-hos; Nova, the <literal>ComputeCapabilitiesFilter</literal>
+  In &kw-hos; &o_comp;, the <literal>ComputeCapabilitiesFilter</literal>
   scheduling filter matches the exact CPU model of the compute host and the
-  requested compute instance model. It doesn't entertain the libvirt's
-  capability of emulating the hosts CPU model to its guests if the guest's CPU
-  model is older or equal to that of the hosts.
+  requested compute instance model. Libvirt supports emulating guest CPUs that
+  are different from the host CPU. However, &o_comp; does not support that
+  feature if the guest's CPU model is older or equal to that of the hosts.
```
sknorr: I don't think this new version makes sense -- I find it rather unlikely that you could emulate a *newer* processor than what you actually have. That would be a pretty extreme case of forward compatibility.

xml/operations-compute-using_flavor_metadata.xml
sknorr: Ah, you are right, I guess. https://wiki.openstack.org/wiki/LibvirtXMLCPUModel

xml/operations-compute-using_flavor_metadata.xml
```diff
@@ -8,154 +8,42 @@
...
+  model may cause &o_comp; to fail with <literal>no valid host found</literal>.
  </para>
  <para>
-  To assist, we have introduced a new Nova scheduler filter which captures the
-  cpu_models as a subset of a particular CPU family. The filter determines if
-  the host CPU model is capable of emulating the guest CPU model by
-  maintaining the mapping of the vCPU models per vendor (currently Intel and
-  AMD) and compares it against the 'model' of the host.
+  To assist, there is a &o_comp; scheduler filter that captures <literal>
+  cpu_models</literal> as a subset of a particular CPU family. The filter
```
sknorr: Don't put a line break directly after the opening tag `<literal>`: That will introduce an ugly leading space which you will see because literals are underlined in our layout.

xml/operations-compute-using_flavor_metadata.xml
```diff
@@ -8,154 +8,42 @@
...
-  scheduling filter matches the exact CPU model of the compute host and the
-  requested compute instance model. It doesn't entertain the libvirt's
-  capability of emulating the hosts CPU model to its guests if the guest's CPU
-  model is older or equal to that of the hosts.
+  In &kw-hos; &o_comp;, the ComputeCapabilitiesFilter checks the exact CPU
+  model of the compute host against the requested compute instance model. It
+  will only pick compute hosts that have the <literal>cpu_model</literal>
+  requested by the instance model, and if the compute host does not have that
+  <literal>cpu_model</literal>, the ComputeCapabilitiesFilter moves on to find
+  another compute host that matches, if possible. Selecting an outdated vCPU
```
sknorr: * You left out the part that libvirt can actually emulate previous-generation CPUs, just Nova doesn't expose that properly in its ComputeCapabilitiesFilter. If that issue is not mentioned, the  new scheduler filter looks a bit out of context.
* Should this word really be "outdated" - what about "unavailable"?

PR #282, opened 2018-05-24
Remove or replace hLinux-bsc#1076357

csymons-suse:
and HPE Linux in all Guides

### DIFF
--- a/xml/operations-cloudadmin_cli.xml
+++ b/xml/operations-cloudadmin_cli.xml
@@ -155,7 +155,7 @@
    Example output:
   </para>
 <screen># This indicates that the admin user is assigned the admin role within the default domain
-ardana@sles:~$ openstack role assignment list --user admin --domain default
+&prompt.ardana; openstack role assignment list --user admin --domain default
 +----------------------------------+----------------------------------+-------+---------+---------+
 | Role                             | User                             | Group | Project | Domain  |
 +----------------------------------+----------------------------------+-------+---------+---------+
@@ -171,7 +171,7 @@ ardana@sles:~$ openstack role assignment list --user admin --domain default
    <literal>test@example.com</literal> and a description of <literal>Test
    User</literal>:
   </para>
-<screen>ardana@sles:~$ openstack user create --domain default --password-prompt --email test@example.com --description "Test User" --enable testuser
+<screen>&prompt.ardana; openstack user create --domain default --password-prompt --email test@example.com --description "Test User" --enable testuser
 User Password:
 Repeat User Password:
 +-------------+----------------------------------+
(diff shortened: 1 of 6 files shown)


xml/operations-cloudadmin_cli.xml
```diff
@@ -23,20 +23,20 @@
...
-    settings are preconfigured in the file <filename>~/keystone.osrc</filename>.
-    By default, <filename>keystone.osrc</filename> is configured with the admin
-    endpoint of keystone. If the admin endpoint is not accessible from your
-    network, change <literal>OS_AUTH_URL</literal> to point to the public
-    endpoint.
+    <emphasis>default</emphasis> domain via the &o_ident; v3 identity API.
+    These settings are preconfigured in the file
+    <filename>~/keystone.osrc</filename>. By default,
+    <filename>keystone.osrc</filename> is configured with the admin endpoint of
+    keystone. If the admin endpoint is not accessible from your network, change
```
dmpop: should *keystone* be *&o_ident;* ?

xml/operations-cloudadmin_cli.xml
```diff
@@ -155,7 +155,7 @@
    Example output:
   </para>
 <screen># This indicates that the admin user is assigned the admin role within the default domain
-stack@hLinux:~$ openstack role assignment list --user admin --domain default
+ardana@sles:~$ openstack role assignment list --user admin --domain default
```
sknorr: I see why you'd want to change these prompts, but this is not really the right way. Take a look at entity-decl.ent, line 20-23 -- this is how we usually do prompts:

* We use the prompt tag which in HTML gets CSS to make it impossible to accidentally copy it (also adds gray [=user] or red [= root] color)
* We do not include the path, because that is often just very long.
* We do not usually include a machine name. (Except for setups where you need to be able to differentiate between machines.)
* We use the SUSE prompt separators (`>`/`#`).
* We normally use an entity for prompts.

Because `ardana` is kind of a special user, I would suggest to create a new `&prompt.ardana;` for user `ardana` and then use that everywhere (i.e. copy line 22 and change the entity name and user name in the prompt).

xml/operations-opsconsole-en-cs_opsconsole_hardware_baremetal.xml
```diff
@@ -36,8 +36,8 @@
   <listitem>
    <para>
     View and provision bare metal servers that are not managed by HPE OneView
```
sknorr: Can you do "HPE OneView"->"\&oneview;" and also list \&susemgr; here?

xml/operations-opsconsole-en-cs_opsconsole_hardware_baremetal.xml
```diff
@@ -36,8 +36,8 @@
...
-    by installing an HPE Linux or RHEL KVM operating system. You can then
-    activate the servers as compute hosts.
+    by installing an &sle_repo; KVM operating system. You can then activate the
```
sknorr: Hm, does HPE still support RHEL here? If so, you should add that, but vendored with vendor="hpe"

PR #285, opened 2018-05-29
Fix the swift container-sync documentation

matthewoliver:
The realm keys are used by container-sync to create a trust
with another cluster in the realm. They _should_ never be used
as the keys used in the container metadata when setting up the
sync between containers.

Container sync will use both the realm key(s) and the key specified
in the container metadata to generate a HMAC to authenticate the
container-sync request to push objects to the destination container.

The doc specifically said to use the realm key when setting up
sync between containers, which is not only incorrect but insecure.
General users should not know the realm keys. Ops should set up the
trust (realm keys). Users can use /info to see other clusters
and will generate their own keys to set up container sync relationships
with containers.

### DIFF
--- a/xml/operations-objectstorage-swift_container_sync.xml
+++ b/xml/operations-objectstorage-swift_container_sync.xml
@@ -328,6 +328,12 @@ Additional middleware: container_sync
 key = lQ8JjuZfO
 # key2 =
 cluster_thiscluster = http://&lt;Swift-Proxy-Vip&gt;:8080/v1/</screen>
+  <para>
+   The keys defined in <literal>/etc/swift/container-sync-realms.conf</literal>
+   are used by the container-sync daemon to determine trust. On top of this
+   the containers that will be in sync will need a seperate shared key they
+   both define in container metadata to establish their trust between each other.
+  </para>
   <orderedlist>
    <listitem>
     <para>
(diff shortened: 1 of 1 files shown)


xml/operations-objectstorage-swift_container_sync.xml
```diff
@@ -328,6 +328,12 @@ Additional middleware: container_sync
 key = lQ8JjuZfO
```
aojea: Should we change this key for 'secret123'  too ?

matthewoliver: No, that one is fine. That's the container-sync realms file. The one Ops set up. so in that case it _is_ the realm key. The one a normal user shouldn't know about.

the sectret123 is the user generated key they add to both containers they want to sync between.

A HMAC is generated on the object, realm key, user key, operation etc, so the other cluster can generate a matching HMAC to see if it's allowed.

xml/operations-objectstorage-swift_container_sync.xml
```diff
@@ -328,6 +328,12 @@ Additional middleware: container_sync
...
 # key2 =
 cluster_thiscluster = http://&lt;Swift-Proxy-Vip&gt;:8080/v1/</screen>
+  <para>
+   The keys defined in <literal>/etc/swift/container-sync-realms.conf</literal>
+   are used by the container-sync daemon to determine trust. On top of this
+   the containers that will be in sync will need a seperate shared key they
+   both define in container meta-data to esptablish their trust between each other.
```
sknorr: * typo: es_p_tablish
* terminology: please use "metadata", not "meta-data"

xml/operations-objectstorage-swift_container_sync.xml
```diff
@@ -357,17 +363,16 @@ Containers in policy "erasure-code-ring": 3
    </listitem>
    <listitem>
     <para>
-     Configure container-src to sync to container-dst using the key specified
-     in the intracluster realm in
-     <literal>/etc/swift/container-sync-realms.conf</literal>
+     Configure container-src to sync to container-dst using a key specified
+     by both containers. In this example we'll use 'secret123'.
```
sknorr: Generally, we use the tag `<replaceable>` (shown in italics in the output) in these cases, for example like:

* `Replace  <replaceableable>KEY</replaceable> with your key.`
* `<screen>swift post ... -k '<replaceable>KEY</replaceable>' ...</screen>`

matthewoliver:
Thanks @sknorr fixed nits

sknorr:
Thanks! Merged now.

(Note that I pushed to your branch to fix a `replaceableable` tag.)

PR #299, opened 2018-06-08
 PCI-PT boot fails. Fix bsc#1084132

dmpop:


swami-dvr:
This sentence can be re-phrased.
with dual-port PCI-PT NICs should always be passed in with both ports to the VM.
'with dual-port PCI-PT NICs, both ports should always be passed in to the VM.'

PR #302, opened 2018-06-09
TR Bosworth Planning Guide bsc#1095280

csymons-suse:
remove what's new section
move Compute Nodes out of Alternative Configurations

### DIFF
--- a/xml/installation-installation-sles-install_sles.xml
+++ b/xml/installation-installation-sles-install_sles.xml
@@ -33,15 +33,6 @@
    </listitem>
    <listitem>
     <para>
-     The contents of the &slsa; SDK ISO (SLE-12-SP3-SDK-DVD-x86_64-GM-DVD1.iso)
-     must be mounted or copied to
-     <literal>/opt/ardana_packager/ardana/sles12/zypper/SDK/</literal>. If you choose
-     to mount the ISO, we recommend creating an <literal>/etc/fstab</literal>
-     entry to ensure the ISO is mounted after a reboot.
-    </para>
-   </listitem>
-   <listitem>
-    <para>
      You must identify the node(s) on which you want to install &slsa;, by adding
      the key/value pair <literal>distro-id: sles12sp3-x86_64</literal> to
      server details in <literal>servers.yml</literal>. You will also need to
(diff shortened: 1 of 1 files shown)


xml/installation-installation-sles-install_sles.xml
```diff
@@ -19,11 +19,8 @@
  <section>
   <title>Deploying legacy BIOS &slsa; compute nodes</title>
   <para>
-   <!-- FIXME: This paragraph is now saying that installing SLES is like
-   installing SLES because we replaced hlinux->SLES. - sknorr, 2018-02-05 -->
-   The installation process for &slsa; nodes is almost identical to that of
-   &hlinux; nodes as described in <xref linkend="install_kvm"/>.
-   The key differences are:
+   The installation process for &slsa; nodes is similar to that described in
+   <xref linkend="install_kvm"/> with the following differences:
```
sknorr: Mmmh, given that the nodes described in install_kvm are *also* SLES 12 SP3 -- why should there be any difference whatsoever? I guess we now have documented this twice (and perhaps one of these is wrong actually) and hence should probably be deduplicated/merged.
(We do still need the RHEL document for HPE in addition *somewhere*.)

xml/planning-planning-compute-node_input_model.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
```
sknorr: If in the end, you do need this new file (see my first comment), please give it a shorter name, e.g. `planning-compute-model.xml` or so (use `git mv OLD_NAME NEW_NAME` to do that). The `planning-planning` stuff comes from dietrich and we somewhat missed the chance to rename the initial files when it was still painless.

xml/installation-installation-sles-install_sles.xml
csymons-suse: I think that they are different. This installation includes an SDK. It has different directory instructions. This file (installation-installation-sles-install_sles.xml) is not related to the RHEL information. That's in the Planning Guide.

If you still think that there's duplication, I can ask some tech person to go through it. I don't see it that way, but I also don't know how these installation processes actually work.

csymons-suse: What I wrote is partially incorrect. bsc# 1089358 (Dirk) says SDK is not needed. Nonetheless, I think that the instructions are different between the two sections. Tell me what you think and I'll get a technical person to have a look if you still think it's messed up.
Part of the Doc Squad is that well-informed people will be looking at the documentation for accuracy...as the first order of business.

sknorr: Yes, I guess a review from someone technical comparing these two sections might be the best option. From what I understand, HPE originally supported HLinux (based on Debian) and later expanded support to include RHEL and SLES. The "install_kvm" section used to document HLinux and this section used to document SLES. Since we replaced HLinux->SLES, supposedly both are documenting SLES.
[Also, in the Installation guide, we have a file installation-installation-rhel-install_rhel.xml as well (vendored HPE-only).]

PR #307, opened 2018-06-13
Migration documentation fixes and updates

cadenzajon:


cadenzajon:
@sknorr Of the docs changes you refer to:
5% were inappropriate and I reverted them
20% had slight errors that I corrected
75% were great and I left as-is

In addition, there are quite a few supplementary updates to migration docs I've made in this review. Also I have now included the results of my MariaDB/MySQL review. Thanks!

PR #317, opened 2018-06-13
SCRD-3598 remove refrence of newton in HPE docs (bsc#1076233)

guglo:
This checkin includes:
1. in HPE operation guide, update keystone configuration link to use git keystone sample
and removed the Newton refrence
2. in HPE user guide, updated mangnum images to use later bits close to pike release
and removed the Newton refrerence


ramak-suse:
Ok

PR #332, opened 2018-06-15
Fix for bsc#1046336

fsundermeyer:
CVE-2016-4383: openstack-glance: Reuse of image IDs may allow malicious users
to replace images for other users

### DIFF
--- a/xml/depl_policy_json.xml
+++ b/xml/depl_policy_json.xml
@@ -343,8 +343,11 @@
      <screen>"deactivate": "role:admin"
 "reactivate": "role:admin"</screen>
      <para>
-      These <emphasis role="bold">settings should not be changed!</emphasis>
-     </para>
+      &suse; suggests these settings should <emphasis>not</emphasis> be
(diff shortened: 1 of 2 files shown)


KeithMnemonic:
I am not sure we want to say something as simple as "these settings should no be changed" . perhaps we need something more like.  "SUSE suggests these settings should not be changed. If you do change them please refer to the OSSN..... for details on the exact scope of the security issue"

xml/depl_policy_json.xml
```diff
@@ -277,4 +277,79 @@
...
+     in <filename>cinder policy.json</filename>. Assign this role to users
+     whose job function it is to perform Cinder storage-related administrative
+     tasks.
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     <literal>glance_admin</literal> role in the identity service and support
+     in <filename>glance policy.json</filename>. Assign this role to users
+     whose job function it is to perform Cinder storage-related administrative
```
bryanstephenson: "Cinder storage" -> "Glance image"

PR #340, opened 2018-06-19
Update Monasca Transform docs bsc#1086974

aagate:
* Update the change hosts filter step to refer
  to a transform_specs.json.j2 template which
  is now managed by CLM.

### DIFF
--- a/xml/operations-monitoring-crt_change_compute_host.xml
+++ b/xml/operations-monitoring-crt_change_compute_host.xml
@@ -18,31 +18,31 @@
  <para>
   <emphasis role="bold">Steps</emphasis>
  </para>
- <orderedlist>
-  <listitem>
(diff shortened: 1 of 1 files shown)


xml/operations-monitoring-crt_change_compute_host.xml
```diff
@@ -56,7 +44,7 @@
   </listitem>
```
fsundermeyer: I cannot comment on lines 33 and 27 directly, so it goes here ;-):

The long lines starting with &lt;screen>{"aggregation_params_map do not break correctly in the PDF. Two suggestions:

* replace "," with ", " (comma+space) to allow line breaks after the comma (AFAIK this is still valid json)
* ommit large parts of the string by replacing it with [...]

PR #391, opened 2018-07-05
Fix start database procedure in disaster recovery test (bsc#1081097)

aagate:
Remove reference to percona-bootstrap playbook and replace with
galera-bootstrap.yml which also starts the db cluster.
Also added a note to  refer to section to manually start
mariadb database if the playbook fails to start db cluster.


aagate:
Thanks @csymons-suse! Changed the link and the wording

PR #395, opened 2018-07-06
remove duplicate para

csymons-suse:
remove duplicate line in Operations - 13.2.1.1 Full Disaster Recovery


sknorr:
Actually, while it looks good, this is a change I also happened to make yesterday. My change was just merged, so this won't be needed anymore.

PR #403, opened 2018-07-16
Update CaaSP 3 image filenames in install guide

r0ckarong:
Also removed duplicate file and updated overview reference


xml/install_caasp_heat_templates.xml
```diff
@@ -105,5 +105,11 @@ CaaSP-3</screen>
     </para>
    </step>
   </procedure>
+  <para>
+   When you have successfully accessed the admin node web interface via the
+   floating IP, follow the instructions at <link
+   xlink:href="https://www.suse.com/documentation/suse-caasp-3/book_caasp_deployment/data/book_caasp_deployment.html"/> to
+   continue the set up of &caasp;.
```
sknorr: "setup" is a noun here, so should be one word.

r0ckarong: That was there from the beginning but I have updated it.

PR #406, opened 2018-07-17
change requested SCRD3709-imagereview6.odt

csymons-suse:
label change in file system diagram


guglo:
@csymons-suse  think we need to talk about this first based on Chris Devita the response.

csymons-suse:
diagram text has been superseded. No longer valid.

PR #434, opened 2018-07-26
Update support matrix. Fix bsc#1076358

dmpop:


csymons-suse:
bug is still NEEDINFO

PR #453, opened 2018-08-07
Updates to migration docs

cadenzajon:


xml/art_migration.xml
```diff
@@ -110,6 +110,22 @@
...
+   <step>
+    <para>
+     Because all Freezer content will be obsolete upon migrating to the new platform,
+     it is recommended to stop Freezer jobs and eliminate historic backups.
+    </para>
+<screen><prompt>(old-deployer)> </prompt>source ~/backup.osrc
+<prompt>(old-deployer)> </prompt>for id in $(freezer job-list -f value | awk '{print $1}')
+  do freezer job-stop $id
+done
+<prompt>(old-deployer)> </prompt>for container in $(openstack container list -f value)
```
kwu83tw: Just curious if Freezer is the only service who reserves container, cause it'll remove all contents within containers.

PR #466, opened 2018-08-13
add first time note bsc#1104699

csymons-suse:
Install UI only works on initial deployments. Note added to current documentation. Will be changed in code in the future.


jeremy-moffitt:
built changeset and confirmed message is present, merging the change

PR #467, opened 2018-08-14
remove ilo from diagram

csymons-suse:


jeremy-moffitt:
change looks good, is there a ticket associated with this?

csymons-suse:
related to bsc#1092569 iLO is HPE specific and should be replaced

PR #473, opened 2018-08-19
add cinder_admin command

csymons-suse:
information is in Security Guide, needs to be in the Installation Guide.

### DIFF
--- a/xml/installation-installation-configure_3par.xml
+++ b/xml/installation-installation-configure_3par.xml
@@ -81,6 +81,12 @@
  <section xml:id="idg-installation-installation-configure_3par-xml-9">
   <title>Notes</title>
   <para>
+   The <literal>cinder_admin</literal> role must be added in order to configure
+   3Par ICSI as a volume type in &o_dash;.
+  </para>
+  <screen>&prompt.ardana;source ~/service.osrc
+&prompt.ardana;openstack role add --user admin --project admin cinder_admin</screen>
+  <para>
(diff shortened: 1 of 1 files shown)


csymons-suse:
This PR can be merged. It fixes the first issue in the bug report. NEEDINFO has been added to the JIRA card and to the bug. Recommend closing this bug and filing another just to cover the second issue--wrong file name.

KeithMnemonic:
The defaults in the cinder.conf have the correct name. This is not a code issue

http://git.suse.provo.cloud/cgit/ardana/cinder-ansible/tree/roles/_CND-CMN/templates/cinder.conf.j2?h=stable/pike


If the DOCS are still referencing the old one that section needs to be updatesd



http://docserv.nue.suse.com/documents/SUSE_OpenStack_Cloud_8/hpe-helion-openstack-installation/single-html/#config_3par

3PAR driver has updated name: In the OpenStack Mitaka release, the 3PAR driver used for HPE Helion OpenStack integration had its name updated from HP3PARFCDriver and HP3PARISCSIDriver to HPE3PARFCDriver and HPE3PARISCSIDriver. To prevent issues when upgrading from previous HPE Helion OpenStack releases, we left the names as-is in the release and provided a mapping so that the integration would continue to work. This will produce a warning in the cinder-volume.log file advising you of the deprecated name. The warning will look similar to this: “Option "hp3par_api_url" from group "<YOUR_SECTION>" is deprecated. Use option "hpe3par_api_url" from group "<YOUR_SECTION>"'”.

These are just warnings and can be ignored.

PR #478, opened 2018-08-23
Overhaul of RHEL compute node documentation, plus migration doc tweaks

cadenzajon:


xml/installation-installation-rhel-install_rhel_compute_node.xml
```diff
@@ -0,0 +1,175 @@
...
+     to other nodes with the following command:
+    </para>
+<screen>sudo ln -s /srv/www/htdocs/repo/RPMMD/CentOS /opt/ardana_packager/ardana/rhel7/yum/centos</screen>
+   </step>
+   <step>
+    <para>
+     Add this new repo as a Yum package source to cloud nodes by populating
+     <filename>/etc/yum.repos.d/ardana-centos.repo</filename> in each
+     &rhla; system with the following contents:
+    </para>
```
KeithMnemonic: we should a note here  like "If using a remote SMT, follow the same steps for this repo as you are doing for the other repos noting the correct target path on the deployer. Please see Chapter 5 on the install guide for details ....

PR #495, opened 2018-09-05
Feature/operations mtce update

csymons-suse:


csymons-suse:
pulled in extraneous commit

PR #511, opened 2018-09-24
Correct Swift region references in Administration Guide (bsc#1105132)

csymons-suse:


xml/bk_openstack_admin.xml
```diff
@@ -13263,9 +13260,7 @@ seeds the Python pseudo-random number generator.</para>
           <step>
             <para>The ring builder assigns each partition replica to the device that
 requires most partitions at that point while keeping it as far away
-as possible from other replicas. The ring builder prefers to assign a
-replica to a device in a region that does not already have a replica.
```
matthewoliver: Why are we removing references to the swift regions? this is true, and a region in Swift isn't necessarily equivalent to regions in keystone or nova.

xml/bk_openstack_admin.xml
```diff
@@ -13206,10 +13206,7 @@ partitions have four replicas and 80 percent have three replicas.</para>
         <para>The replica count is adjustable. For example:</para>
         <screen>$ swift-ring-builder account.builder set_replicas 4
 $ swift-ring-builder account.builder rebalance</screen>
-        <para>You must rebalance the replica ring in globally distributed clusters.
-Operators of these clusters generally want an equal number of replicas
-and regions. Therefore, when an operator adds or removes a region, the
-operator adds or removes a replica. Removing unneeded replicas saves on
+        <para>Removing unneeded replicas saves on
 the cost of disks.</para>
```
matthewoliver: removing replicas do save on cost of disks but you loose durability.

csymons-suse: Please suggest alternative content for removing/not removing replicas.

xml/bk_openstack_admin.xml
csymons-suse: Thank you for the comments. Would you please suggest more accurate content?

PR #535, opened 2018-09-28
Revise migration documentation based on further testing (bsc#1109018)

cadenzajon:


xml/art_migration.xml
```diff
@@ -395,8 +396,20 @@ done</screen>
...
+       In Distributed Virtual Router (DVR) enabled clouds, responsibility for
+       network routing is shared across the compute nodes, and downtime of the
+       seed node as a controller will have no adverse effect on workload network
+       traffic. To see if DVR is enabled in your cloud, run the following command
+       and check the entries in the "Distributed" column.
+      </para>
+<screen><prompt>(old-deployer)> </prompt>source ~/service.osrc
+<prompt>(old-deployer)> </prompt>neutron router-list</screen>
+      <para>
+       If you observe non-distributed routers, then the next step will ensure
```
bryanstephenson: I don't have access to a CVR cloud to check, but is it possible to qualify the statement "If you observe non-distributed routers" more, for example to "If you observe non-distributed routers running on the seed node"? If so we should do that. Looks good otherwise.

xml/art_migration.xml
```diff
@@ -395,8 +396,20 @@ done</screen>
...
+       that any responsibility for those routers' function is migrated away from
+       the seed node to other nodes in the controller cluster, so that routing
+       remains available even while the seed node is taken offline. (This step
+       is not necessary if all your routers are distributed.)
```
bryanstephenson: Similarly, could we say "if all your routers are distributed or no Centralized Virtual Routers are running on the seed node" ?

xml/art_migration.xml
cadenzajon: To me, it looks like the output of "neutron router-list" does not indicate to the user which controller node a CVR router is running on. So, it's safest to run the evacuation script any time you see non-distributed routers (which should be a no-op if the planned seed node is not hosting any CVR routers anyway).

xml/art_migration.xml
cadenzajon: See above, I prefer preserving the current content.

csymons-suse:
superseded by
https://github.com/SUSE-Cloud/doc-cloud/pull/549
with Jonathan Brownell's approval

cadenzajon:
superseded by Carl as described

PR #536, opened 2018-10-01
Feature/soc7 soc8 postgre mariadb upgr

dmpop:
Add PostgreSQL to MariaDB migration instructions to the upgrade chapter.

### DIFF
--- a/xml/depl_maintenance.xml
+++ b/xml/depl_maintenance.xml
@@ -621,7 +621,7 @@ Error during prechecks. 4 Byte UTF8 characters found in the source database.</sc
       <para>
        Stop chef-client services on the nodes, to prevent regular runs of
        chef-client from starting database-related &ostack; services again. To
-       do this, you can use the <command>knife ssh roles:dns-client systemctl
+       do this from the &admnode;, you can use the <command>knife ssh roles:dns-client systemctl
        stop chef-client</command> command. Stop
        all &ostack; services that make use of the database to prevent them from
        writing new data during the migration.
@@ -672,7 +672,7 @@ migrate</screen>
        <literal>--batch</literal> option and the
        <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
       </para>
-<screen>&prompt.user;psql2mysql.py --batch /etc/pg2mysql/databases.yaml migrate</screen>
+<screen>&prompt.user;psql2mysql --batch /etc/pg2mysql/databases.yaml migrate</screen>
      </step>
      <step>
       <para>
@@ -749,13 +749,11 @@ migrate</screen>
    </sect3>
    <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario2">
     <title>HA Cluster with 2 &contrnode;s</title>
-    <procedure>
-     <step>
      <para>
-     Extend the 2-node cluster with additional node that without any role assigned to
+     Before your proceed, extend the 2-node cluster with additional node that without any role assigned to
      it. Make sure that the new node has enough memory to serve as a &contrnode;.
      </para>
-    </step>
+    <procedure>
     <step>
      <para>
       In &crow; &wi;, switch to the <guimenu>&pacemaker;</guimenu> &barcl;,
--- a/xml/depl_maintenance.xml
+++ b/xml/depl_maintenance.xml
@@ -554,9 +554,36 @@
       <para>
        After the script is finished, you'll find a list of all databases and
        URLs that are ready for data migration in the
-       <filename>/etc/pg2mysql/databases.yaml</filename>. The output of the
-       script contains information on which &dbnode; the file is located.
+       <filename>/etc/pg2mysql/databases.yaml</filename>. The script's output
+      may look as follows:</para>
+      <screen>Preparing node d52-54-77-77-01-01.vo6.cloud.suse.de
+Adding recipe[database::pg2mariadb_preparation] to run_list
+Running chef-client on d52-54-77-77-01-01.vo6.cloud.suse.de...
+Log: /var/log/crowbar/db-prepare.chef-client.log on
+d52-54-77-77-01-01.vo6.cloud.suse.de Run time: 444.725193199s
+Removing recipe[database::pg2mariadb_preparation] from run_list
+Prepare completed for d52-54-77-77-01-01.vo6.cloud.suse.de
+Summary of used databases: /etc/pg2mysql/databases.yaml on
+d52-54-77-77-01-01.vo6.cloud.suse.de</screen>
+      <para>The <literal>Summary of used databases:</literal> line shows the
+      exact location of the <filename>/etc/pg2mysql/databases.yaml</filename> file.
       </para>
+      <para>
+       The <filename>/etc/pg2mysql/databases.yaml</filename> file contains a
+       list of databases along with their source and target connection strings:
+      </para>
+<screen>keystone:
+  source: postgresql://keystone:vZn3nfxXzv97@192.168.243.87/keystone
+  target:
+mysql+pymysql://keystone:vZn3nfxXzv97@192.168.243.88/keystone?charset=utf8
+glance:
+  source: postgresql://glance:cOau7NhaA54N@192.168.243.87/glance
+  target:
+mysql+pymysql://glance:cOau7NhaA54N@192.168.243.88/glance?charset=utf8
+cinder:
+  source: postgresql://cinder:idRll2gJPodv@192.168.243.87/cinder
+  target:
+mysql+pymysql://cinder:idRll2gJPodv@192.168.243.88/cinder?charset=utf8</screen>
      </step>
      <step>
       <para>
@@ -571,7 +598,7 @@
        file for the <systemitem>psql2mysql</systemitem> tool.
       </para>
       <para>
-       Running the tool on the &admnode; makes sense if the input batch file
+       Running the tool on the &admnode; is advisable if the input batch file
        is already located there, or if you prefer not to install additional
        software on cluster nodes.
       </para>
@@ -580,14 +607,105 @@
       <para>
        To determine whether the &postgres; databases contain data that cannot
        be migrated to &mariadb;, run <command>psql2mysql</command> with the
-      <literal>recheck</literal> option:</para>
-      <screen>psql2mysql.py --source $(sed -n -e 's/^.*\(postgresql:\)/\1/p' \
-      /etc/neutron/neutron.conf.d/*) precheck</screen>
+      <literal>precheck</literal> option:</para>
+<screen>psql2mysql \
+--source postgresql://keystone:vZn3nfxXzv97@192.168.243.87/keystone \
+--target mysql+pymysql://keystone:vZn3nfxXzv97@192.168.243.88/keystone?charset=utf8 \
+precheck</screen>
       <para>
-       You can use the <filename>/etc/pg2mysql/databases.yaml</filename> file
-       as the source and input <!-- ??? -->
+       To run precheck operation on all databases in a single operation, use the
+       <literal>--batch</literal> option and the
+       <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
       </para>
+<screen>psql2mysql --batch /etc/pg2mysql/databases.yaml precheck</screen>
       <para>
+       If the precheck indicates that there is data that cannot be imported
+       into &mariadb;, modify the offending data manually to fix
+       the problems.
+      </para>
+     </step>
+     <step>
+      <para>
+       Stop all &ostack; services that make use of the database to prevent them
+       from writing new data during the migration. Stop shef-client services on
+       the nodes, to prevent regular runs from starting database-related
+       &ostack; services again. <emphasis>From this point, &ostack; services
+       become unavailable.</emphasis>
+      </para>
+     </step>
+     <step>
+      <para>
+       You can now migrate databases using the psql2mysql tool. To upgrade a
+       single database, use the following command format:
+      </para>
+<screen>psql2mysql \
+--source postgresql://neutron:secret@192.168.1.1/neutron \
+--target mysql+pymysql://neutron:evenmoresecret@192.168.1.2/neutron?charset=utf8 \
+migrate</screen>
+      <para>
+       To migrate all databases in one operation, use the
+       <literal>--batch</literal> option and the
+       <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
+      </para>
+<screen>psql2mysql.py --batch /etc/pg2mysql/databases.yaml migrate</screen>
+     </step>
+     <step>
+      <para>
+       In the &crow; &wi;, switch to the <guimenu>Database</guimenu>
+       &barcl;. Enable the raw view and set the value of
+       <literal>sql_engine</literal> to <literal>mysql</literal>. Apply the
+       &barcl;. After this step, &ostack; services should be running again and
+       reconfigured to use the &mariadb; database back end.
+      </para>
+     </step>
+     <step>
+      <para>
+        To prevent the &postgres;-related chef code from running, unassign the values
+	from <literal>database-server</literal> role in the <guimenu>Database</guimenu>
+       &barcl;, and apply the &barcl;.
+      </para>
+     </step>
+     <step>
+      <para>
+       Start chef-client services on the nodes again.
+      </para>
+     </step>
+     <step>
+      <para>
+       Stop &postgres; on the &dbnode;s. Uninstall &postgre;packages.
+      </para>
+      <substeps>
+       <step>
+	<para>
+	 To stop the <literal>postgresql</literal> service, run the following
+	 command on each cluster node:
+	</para>
+<screen>crm resource stop postgresql
+crm resource stop fs-postgresql
+crm resource stop drbd-postgresql</screen>
+        <para>
+	 Run the last command only if the previous setup used DRBD.
+	</para>
+       </step>
+       <step>
+	<para>
+	 Remove the packages on all cluster nodes:
+	</para>
+	<screen>zypper rm postgresql94 postgresql94-server</screen>
+       </step>
+       <step>
+	<para>
+	 If you choose not to upgrade to &productname; 8 right away, delete unused pacemaker resource from each cluster node:
+	</para>
+<screen>crm conf delete drbd-postgresql
+crm conf delete fs-postgresql
+crm conf delete postgresql</screen>
+       </step>
+      </substeps>
+     </step>
+     <step>
+      <para>
+       
       </para>
      </step>
     </procedure>
--- a/xml/depl_maintenance.xml
+++ b/xml/depl_maintenance.xml
@@ -636,6 +636,12 @@ Error during prechecks. 4 Byte UTF8 characters found in the source database.</sc
        required downtime. When you perform the actual migration, the data in
        the target databases will be replaced with the latest one in the source databases.
        </para>
+       <para>
+	After the test migration and before the actual migration, it is recommended to run the
+	<command>psql2mysql migrate</command> command to purge tables in the
+	target database. While this step is optional, it speeds up the
+	migration process.
+       </para>
       </note>
       <para>
        On an HA setup, shut down all services that make use of the
@@ -643,13 +649,13 @@ Error during prechecks. 4 Byte UTF8 characters found in the source database.</sc
        example:
       </para>
 <screen>crm resource stop apache2 keystone cinder-api glance-api \
-      neutron-server swift-proxy nova-api magnum-api sahara-api heat-api</screen>   
+      neutron-server swift-proxy nova-api magnum-api sahara-api heat-api ceilometer-collector</screen>   
       <note>
        <para>
 	If the <literal>Manage stateless active/active services with
 	Pacemaker</literal> option in the &pacemaker; &barcl; is set to
 	<literal>false</literal>, the &ostack; services must be stopped on each
-	node using the <command>systemctl</command> command.
+	cluster node using the <command>systemctl</command> command.
        </para>
       </note>
       <para>
@@ -715,7 +721,7 @@ migrate</screen>
 	<para>
 	 Remove the packages on all cluster nodes:
 	</para>
-	<screen>zypper rm postgresql94 postgresql94-server</screen>
+	<screen>&prompt.root;zypper rm postgresql94 postgresql94-server</screen>
        </step>
        <step>
 	<para>
@@ -724,6 +730,14 @@ migrate</screen>
 <screen>&prompt.user;crm conf delete drbd-postgresql
 &prompt.user;crm conf delete fs-postgresql
 &prompt.user;crm conf delete postgresql</screen>
+
+<note>
+ <para>
+  Run the <command>crm conf delete drbd-postgresql</command> command only
+  if the cloud setup your are upgrading uses DRBD.
+ </para>
+</note>
+
        </step>
       </substeps>
      </step>
--- a/xml/depl_maintenance.xml
+++ b/xml/depl_maintenance.xml
@@ -513,17 +513,14 @@
    <para>
     Upgrading &productname; 7 is only possible when it uses &mariadb; deployed
     with the Database &barcl;. This means that before you can proceed with
-    upgrading &productname; 7, you must migrate &postgres; to &mariadb; first.
+    upgrading &productname; 7, you must migrate &postgres; to &mariadb; first. The following description covers several possible scenarios.
    </para>
-   <para>
-    The following description covers several possible scenarios.
-   </para>
-   <sect3>
-    <title>Non-HA Setup. HA Setup with &postgres; and without DRDB</title>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario1">
+    <title>Non-HA setup or HA Setup with a 2-Node Cluster and
+    &postgres; Database Backend</title>
     <para>
      Before you proceed, make sure that the database cluster has an odd number
-     of nodes. If necessary, extend the cluster as described in <!-- FIXME
-     -->. Install the latest maintenance updates on &productname; 7. In the
+     of nodes. If necessary, extend the cluster as described in <xref linkend="sec.postgre-mariadb.upgrade.scenario2"/>. Install the latest maintenance updates on &productname; 7. In the
      &crow; &wi;, switch to the <guimenu>Database</guimenu> &barcl;. You should
      see the new <systemitem>mysql-server</systemitem> role in the
      <guimenu>Deployment</guimenu> section. Add your &dbnode; to the
@@ -532,14 +529,15 @@
      as a back end for &ostack; services. Do not change the
      <literal>sql_engine</literal> at this point.
     </para>
-    <para>
+    <!-- FIXME: Currently doesn't work accoring to devs
+	 <para>
      If &rabbit; is using DRBD, enable the <guimenu>Use RabbitMQ
      Clustering</guimenu> option in the &rabbit; &barcl; under <guimenu>High Availability</guimenu>. Since DRBD is not available
      in &productname; 8, and &rabbit; clustering is supported in &productname;
      7, you should switch to this option now.
-    </para>
+    </para> -->
     <para>
-     Follow the <xref linkend="postgre.mariadb.data.migration"/> procedure to migrate the data from &postgres; to &mariadb;.
+     Follow <xref linkend="postgre.mariadb.data.migration"/> to migrate the data from &postgres; to &mariadb;.
     </para>
     <procedure xml:id="postgre.mariadb.data.migration">
      <title>Data Migration</title>
@@ -672,7 +670,7 @@ migrate</screen>
      </step>
      <step>
       <para>
-       Stop &postgres; on the &dbnode;s. Uninstall &postgre;packages.
+       Stop &postgres; on the &dbnode;s. Uninstall &postgres;packages.
       </para>
       <substeps>
        <step>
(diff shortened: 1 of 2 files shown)


xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
 
+  <sect2 xml:id="sec.depl.maintenance.postgre-mariadb.upgrade">
+   <title>Preparing &postgres;-Based &productname; 7 for Upgrade</title>
+   <para>
+    Upgrading &productname; 7 is only possible when it uses &mariadb; deployed
+    with the Database &barcl;. This means that before you can proceed with
+    upgrading &productname; 7, you must migrate &postgres; to &mariadb; first. The following description covers several possible scenarios.
+   </para>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario1">
+    <title>Non-HA setup or HA Setup with a 2-Node Cluster and
```
jsuchome: Oh, here has to be "more than 2 node cluster". Or just Database cluster with odd number of nodes.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+    &postgres; Database Backend</title>
+    <para>
+     Before you proceed, make sure that the database cluster has an odd number
+     of nodes. If necessary, extend the cluster as described in <xref linkend="sec.postgre-mariadb.upgrade.scenario2"/>. Install the latest maintenance updates on &productname; 7. In the
+     &crow; &wi;, switch to the <guimenu>Database</guimenu> &barcl;. You should
```
jsuchome: I believe there are actually 2 sections describing extending the cluster (one with DRBD one without).

(It might be better to merge them into one and make the DRBD part just an extra note in such section)

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+     see the new <systemitem>mysql-server</systemitem> role in the
+     <guimenu>Deployment</guimenu> section. Add your &dbnode; to the
+     <systemitem>mysql-server</systemitem> role and apply the
+     &barcl;. &mariadb; is now deployed and running, but it is still not used
+     as a back end for &ostack; services. Do not change the
+     <literal>sql_engine</literal> at this point.
```
jsuchome: This line ("do not change") fits better before the one about applying the proposal

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+           Run the <systemitem>/opt/dell/bin/prepare-mariadb</systemitem>
+	   script on the &admnode; to prepare the &mariadb; instance by creating
+	   the required users, databases, and tables.
+      </para>
+     </step>
+     <step>
+      <para>
+       After the script is finished, you'll find a list of all databases and
+       URLs that are ready for data migration in the
+       <filename>/etc/pg2mysql/databases.yaml</filename>. The script's output
```
jsuchome: The yaml file will be located "at one of the database nodes."

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      exact location of the <filename>/etc/pg2mysql/databases.yaml</filename> file.
+      </para>
+      <para>
+       The <filename>/etc/pg2mysql/databases.yaml</filename> file contains a
+       list of databases along with their source and target connection strings:
+      </para>
+<screen>keystone:
+  source: postgresql://keystone:vZn3nfxXzv97@192.168.243.87/keystone
+  target:
+mysql+pymysql://keystone:vZn3nfxXzv97@192.168.243.88/keystone?charset=utf8
```
jsuchome: This seems somewhat badly formatted. Could you move the "mysql+pymysql" line right after `target: ` ?

And if not, consider just using some fake password to make the strings shorted.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+cinder:
+  source: postgresql://cinder:idRll2gJPodv@192.168.243.87/cinder
+  target:
+mysql+pymysql://cinder:idRll2gJPodv@192.168.243.88/cinder?charset=utf8</screen>
+     </step>
+     <step>
+      <para>
+       Install the <systemitem>psql2mysql</systemitem> migration tool (it
+       is a part of <package>python-psql2mysql</package> package) on the
+       &admnode; or directly on the &dbnode;.
```
jsuchome: OK, let's just say, install "python-psql2mysql" package

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      </para>
+      <para>
+       Running the tool directly on &dbnode; speeds up the migration operation,
+       as this does not require moving database data over the network. You can
+       also use <filename>/etc/pg2mysql/databases.yaml</filename> as the input
+       file for the <systemitem>psql2mysql</systemitem> tool.
```
jsuchome: I think you can add that this is the recommended way.

xml/depl_maintenance.xml
dmpop: But in this case, the first two sentences don't make sense:

*Before you proceed, make sure that the database cluster has an odd number of nodes. If necessary, extend the cluster as described in [...]

Or am I missing something?

xml/depl_maintenance.xml
jsuchome: Also, I kind of think we should mention to curious users there's a README file for the tool. It should be installed to /usr/share/doc/packages/python-psql2mysql/README.md

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      <para>
+       If the precheck indicates that there is data that cannot be imported
+       into &mariadb;, modify the offending data manually to fix
+       the problems.
+      </para>
+     </step>
+     <step>
+      <para>
+       Stop all &ostack; services that make use of the database to prevent them
+       from writing new data during the migration. Stop shef-client services on
```
jsuchome: Typo: `shef` -> `chef`

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+       the nodes, to prevent regular runs from starting database-related
```
jsuchome: I would say "regular runs of _chef-client_"

xml/depl_maintenance.xml
dmpop: This is a bit confusing. Could you elaborate on that, please?

xml/depl_maintenance.xml
jsuchome: For HA based setups, remember to shut down the services using `crm` command, e.g. `crm resource stop nova-api`

xml/depl_maintenance.xml
dmpop: I dropped this, because the text on line 568 explains that the Summary of used databases line shows the exact location of the /etc/pg2mysql/databases.yaml file.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

jsuchome: Also: **First**, stop chef-client services. **Then** OpenStack ones

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      </para>
+     </step>
+     <step>
+      <para>
+       Start chef-client services on the nodes again.
+      </para>
+     </step>
+     <step>
+      <para>
+       Stop &postgres; on the &dbnode;s. Uninstall &postgres;packages.
```
jsuchome: There's a space missing between "&postgres;" and packages

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      </para>
+      <substeps>
+       <step>
+	<para>
+	 To stop the <literal>postgresql</literal> service, run the following
+	 command on each cluster node:
```
jsuchome: Not on each, just on _one_ cluster node

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+	</para>
+<screen>crm conf delete drbd-postgresql
+crm conf delete fs-postgresql
+crm conf delete postgresql</screen>
+       </step>
+      </substeps>
+     </step>
+     <step>
+      <para>
+       While you can continue using DRBD with &rabbit;, it is recommended to
```
jsuchome: Ah, we have to change this.
As mentioned earlier, RabbitMQ switch does not work before the upgrade. So we must write something like:

"If DRBD is not used as a backend for RabbitMQ, it is possible to remove it at this point. Delete the packages with zypper and claim the DRBD disks if you want. Otherwise, leave this step until after the full upgrade is done. RabbitMQ setup will be switched from DRBD during the upgrade procedure".

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+       switch to the clustering setup. If you don't make the switch at this
+       point, the upgrade to &productname; 8 will do this automatically. If you
+       choose to switch to the clustering setup, stop DRBD and uninstall
+       DRBD-related packages using the following command:
+      </para>
+<screen>&prompt.sudo;zypper rm drbd drbd-utils</screen>
+     </step>
+     <step>
+      <para>
+       Reclaim the disk space sued by &crow; for DRBD. To do this, edit the node data using <systemitem>knife</systemitem>:
```
jsuchome: I think this should be part of previous step.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+      <para>
+        Search for <literal>claimed_disks</literal> and remove the entry with owner set to <literal>LVM_DRBD</literal>.
+      </para>
+     </step>
+    </procedure>
+   </sect3>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario2">
+    <title>HA Setup with Two &contrnode;s without DRBD</title>
+    <para>
+     Extend the cluster with additional node that without any role assigned to
```
jsuchome: _additional node that does not have any other role assigned yet_ ?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,262 @@
...
+     it. In &crow; &wi;, switch to the <guimenu>&pacemaker;</guimenu> &barcl;,
+     add the <literal>pacemaker-cluster-member</literal> role to the new node,
+     and apply &barcl;. Proceed with the migration procedure as described in
+     <xref linkend="sec.postgre-mariadb.upgrade.scenario1"/>.
+    </para>
+   </sect3>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario3">
+    <title>HA Setup with 2 &contrnode;s and DRBD Used for Database or &rabbit;</title>
```
jsuchome: Please merge with the previous section and only mention that point 2 ( allow_larger_cluster part) is specific to DRBD setup

xml/depl_maintenance.xml
dmpop: Ah, got it! Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed. Doesn't look like the `resource stop` command is covered in the HA guide, though.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
jsuchome: I'd remove the _Before you proceed, make sure that the database cluster has an odd number of nodes. If necessary, extend the cluster as described in_ part.

That should be covered in other sections and this one is only relevant once user has more than 2 nodes in a cluster.

BTW we're talking about database cluster. If user has more clusters (like for network or OpenStack services), we do not care about them here

xml/depl_maintenance.xml
jsuchome: BTW there's a typo here (sued -> used)

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
```diff
@@ -703,37 +711,25 @@ crm conf delete postgresql</screen>
...
-       choose to switch to the clustering setup, stop DRBD and uninstall
-       DRBD-related packages using the following command:
+       If DRBD is not used as a backend for &rabbit;, it is possible to remove it at this point, using the following command:
       </para>
 <screen>&prompt.sudo;zypper rm drbd drbd-utils</screen>
-     </step>
-     <step>
       <para>
-       Reclaim the disk space sued by &crow; for DRBD. To do this, edit the node data using <systemitem>knife</systemitem>:
+       You can then reclaim the disk space sued by &crow; for DRBD. To do this, edit the node data using <systemitem>knife</systemitem>:
```
jsuchome: Still the _sued_ typo

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+       <literal>--batch</literal> option and the
+       <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
+      </para>
+<screen>psql2mysql --batch /etc/pg2mysql/databases.yaml precheck</screen>
+      <para>
+       If the precheck indicates that there is data that cannot be imported
+       into &mariadb;, modify the offending data manually to fix
+       the problems.
+      </para>
+     </step>
```
jsuchome: OK, so let's make it little bit more complicated (but also better for some users).

It should be stated that users can run `psql2mysql migrate` command even before stopping the services, just for testing. Subsequent runs of migrate will always replace data in target databases with latest state in source one.
So user can run `migrate` and if something fails (that wasn't catched with `precheck`), there's still a chance to fix the problems _before_ the OpenStack services are down.

But once they want to proceed with the real migration, yes, services need to be down.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
 
+  <sect2 xml:id="sec.depl.maintenance.postgre-mariadb.upgrade">
+   <title>Preparing &postgres;-Based &productname; 7 for Upgrade</title>
+   <para>
+    Upgrading &productname; 7 is only possible when it uses &mariadb; deployed
+    with the Database &barcl;. This means that before you can proceed with
+    upgrading &productname; 7, you must migrate &postgres; to &mariadb; first. The following description covers several possible scenarios.
+   </para>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario1">
+    <title>Non-HA setup or HA Setup with a 2-Node Cluster and
```
rhafer: Hm, why is this mentioning 2-Node postgresql clusters explicitly? The instructions below are perfectly fine for 3-node postgresql clusters as well, no? (And there is no section covering >2 node clusters AFAICS)

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+    &postgres; Database Backend</title>
+    <para>
+     Before you proceed, make sure that the database cluster has an odd number
+     of nodes. If necessary, extend the cluster as described in <xref
+     linkend="sec.postgre-mariadb.upgrade.scenario2"/>. Install the latest
+     maintenance updates on &productname; 7. In the
+     &crow; &wi;, switch to the <guimenu>Database</guimenu> &barcl;. You should
+     see the new <systemitem>mysql-server</systemitem> role in the
+     <guimenu>Deployment</guimenu> section. Do not change the
+     <literal>sql_engine</literal> at this point. Add your &dbnode; to the
```
rhafer: Assuming that `&dbnode` expands to "database node" I think this is a bit misleading. In the HA case you won't assign and single node to the role, but a whole cluster.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+  target: mysql+pymysql://cinder:idRll2gJPodv@192.168.243.88/cinder?charset=utf8</screen>
+     </step>
+     <step>
+      <para>
+       Install the <package>python-psql2mysql</package> package on the
+       &admnode; or directly on the &dbnode;.
+      </para>
+      <para>
+       Running the tool directly on &dbnode; speeds up the migration operation,
+       as this does not require moving database data over the network. You can
```
rhafer: Hm, this is probably a bit too simplified. The migration will still require network traffic (at the bare mininum for the galera replication). However running it on a controller node might reduce the traffice if that e.g. is the currently (active) postgresql node. Ideally we would make sure to have the active postgresql, haproxy and mariadb node on the same machine. But there is currently nothing in the tools that really ensure this.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+      <para>
+       To run precheck operation on all databases in a single operation, use the
+       <literal>--batch</literal> option and the
+       <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
+      </para>
+<screen>psql2mysql --batch /etc/pg2mysql/databases.yaml precheck</screen>
+      <para>
+       If the precheck indicates that there is data that cannot be imported
+       into &mariadb;, modify the offending data manually to fix
+       the problems.
```
rhafer: Should we add screenshot here to show how it looks if such an issue appears?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+      </para>
+     </step>
+     <step>
+      <para>
+       Stop chef-client services on the nodes, to prevent regular runs of
+       chef-client from starting database-related &ostack; services again. Stop
+       all &ostack; services that make use of the database to prevent them from
+       writing new data during the migration.
```
rhafer: Should we add a hint here on how to actually achieve that without manually ssh-ing into every node?

`knife ssh exec roles:dns-client systemctl stop chef-client`

Should actually work. Provided that all openstack nodes have the dns-client node deploy (which should be the case)

xml/depl_maintenance.xml
jsuchome: Rigth, this section should be called "Setup with more than 2 nodes". I think we've already cleared that, but seems like something got lost

xml/depl_maintenance.xml
jsuchome: So maybe we should just tell them to run it on database node and do not go into details?

xml/depl_maintenance.xml
jsuchome: Good idea. I'll try to create such output

jsuchome:
```# psql2mysql --source postgresql://cinder:idRll2gJPodv@192.168.243.86/cinder precheck
Table 'volumes' contains 4 Byte UTF8 characters which are incompatible with the 'utf8' encoding used by MariaDB
The following rows are affected:
+-----------------------------------------+-----------------+-------+
|               Primary Key               | Affected Column | Value |
+-----------------------------------------+-----------------+-------+
| id=5c6b0274-d18d-4153-9fda-ef3d74ab4500 |   display_name  |   💫   |
+-----------------------------------------+-----------------+-------+
Error during prechecks. 4 Byte UTF8 characters found in the source database.```

File: xml/depl_maintenance.xml (hunk 22 above)
Comment (jsuchome) [reply xml/depl_maintenance.xml]: Cool! I did not know (or completely forgot) about `knife ssh`!

Comment (jsuchome) [reply xml/depl_maintenance.xml]: @rhafer It's only `knife ssh roles:dns-client systemctl stop chef-client`, isn't it?

File: xml/depl_maintenance.xml [hunk 23]
```diff
@@ -508,6 +508,258 @@
...
+      </para>
+      <para>
+       On an HA setup, shut down all services that make use of the
+       database. To do this, use the <command>crm</command> command for
+       example:
+      </para>
+<screen>crm resource stop nova-api</screen>
```
jsuchome: We might need to be more specific here and tell users exactly what to stop...

jsuchome: `crm resource stop $resource` where `$resource` is

`apache2
keystone
cinder-api
glance-api
neutron-server
swift-proxy
nova-api`

To be continued...

xml/depl_maintenance.xml
rhafer: Yeah that `exec` was wrong.

xml/depl_maintenance.xml
jsuchome: @rhafer do you know which ceilometer, aodh and barbican services to stop?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,258 @@
...
+     </step>
+    </procedure>
+   </sect3>
+   <sect3 xml:id="sec.postgre-mariadb.upgrade.scenario2">
+    <title>HA Cluster with 2 &contrnode;s</title>
+    <procedure>
+     <step>
+     <para>
+     Extend the 2-node cluster with additional node that without any role assigned to
+     it.
```
jsuchome: Make sure the new node has enough memory to serve as a controller node.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,277 @@
...
+| id=5c6b0274-d18d-4153-9fda-ef3d74ab4500 |   display_name  |   💫   |
++-----------------------------------------+-----------------+-------+
+Error during prechecks. 4 Byte UTF8 characters found in the source database.</screen>
+     </step>
+     <step>
+      <para>
+       Stop chef-client services on the nodes, to prevent regular runs of
+       chef-client from starting database-related &ostack; services again. To
+       do this, you can use the <command>knife ssh roles:dns-client systemctl
+       stop chef-client</command> command. Stop
```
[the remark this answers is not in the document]
dmpop: Fixed.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,277 @@
...
+<screen>&prompt.user;psql2mysql \
+--source postgresql://neutron:secret@192.168.1.1/neutron \
+--target mysql+pymysql://neutron:evenmoresecret@192.168.1.2/neutron?charset=utf8 \
+migrate</screen>
+      <para>
+       To migrate all databases in one operation, use the
+       <literal>--batch</literal> option and the
+       <filename>/etc/pg2mysql/databases.yaml</filename> file as follows:
+      </para>
+<screen>&prompt.user;psql2mysql.py --batch /etc/pg2mysql/databases.yaml migrate</screen>
```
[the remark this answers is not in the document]
dmpop: Fixed.

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,277 @@
...
+       the target databases will be replaced with the latest one in the source databases.
+       </para>
+      </note>
+      <para>
+       On an HA setup, shut down all services that make use of the
+       database. To do this, use the <command>crm</command> command for
+       example:
+      </para>
+<screen>crm resource stop apache2 keystone cinder-api glance-api \
+      neutron-server swift-proxy nova-api magnum-api sahara-api heat-api</screen>
```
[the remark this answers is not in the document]
dmpop: And the name of the ceilometer service is _ceilometer-api_, right?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,277 @@
...
+<screen>&prompt.user;crm resource stop postgresql
+&prompt.user;crm resource stop fs-postgresql
+&prompt.user;crm resource stop drbd-postgresql</screen>
+        <para>
+	 Run the last command only if the previous setup used DRBD.
+	</para>
+       </step>
+       <step>
+	<para>
+	 Remove the packages on all cluster nodes:
```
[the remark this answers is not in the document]
dmpop: Just let me know what I should write there. :-)

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,277 @@
...
+	</para>
+	<screen>zypper rm postgresql94 postgresql94-server</screen>
+       </step>
+       <step>
+	<para>
+	 If you choose not to upgrade to &productname; 8 right away, delete unused pacemaker resource from one cluster node:
+	</para>
+<screen>&prompt.user;crm conf delete drbd-postgresql
```
matthewoliver: Do we need to make a note that you only need to delete this drbd resource if the preious setup used drbd (because we pointed it out in 10a?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,275 @@
...
+      <note>
+       <title>Testing Migration Procedure</title>
+       <para>
+	If you want to perform a dry run of the migration procedure, you can run the
+       <command>psql2mysql migrate</command> without stopping the
+       database-related &ostack; services. This way, if the test migration
+       fails due to errors that weren't caught by the precheck procedure, you
+       can fix them with &ostack; services still running, thus minimizing the
+       required downtime. When you perform the actual migration, the data in
+       the target databases will be replaced with the latest one in the source databases.
```
jsuchome: Please mention also the option to use the `purge-tables` command. After such test migration, we advice users to purge the tables in target database before running the real migration, just to speed up the real migration. It's not necessary, data would be deleted anyway, but it will save some time.

See https://github.com/SUSE/psql2mysql/blob/master/README.md and search for purge-tables

xml/depl_maintenance.xml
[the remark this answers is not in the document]
jsuchome: There's no `ceilometer-api` now. I'm however not sure what to stop, probably `ceilometer-collector`?

xml/depl_maintenance.xml
```diff
@@ -508,6 +508,275 @@
...
+       example:
+      </para>
+<screen>crm resource stop apache2 keystone cinder-api glance-api \
+      neutron-server swift-proxy nova-api magnum-api sahara-api heat-api</screen>   
+      <note>
+       <para>
+	If the <literal>Manage stateless active/active services with
+	Pacemaker</literal> option in the &pacemaker; &barcl; is set to
+	<literal>false</literal>, the &ostack; services must be stopped on each
+	node using the <command>systemctl</command> command.
```
jsuchome: each _cluster_ node

xml/depl_maintenance.xml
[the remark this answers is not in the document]
dmpop: I'm inclined to keep thing simple. So, if there are no objections, let's just keep `zypper rm postgresql94 postgresql94-server`

xml/depl_maintenance.xml
dmpop: Fixed.

xml/depl_maintenance.xml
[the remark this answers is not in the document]
jsuchome: I'm fine with that

jsuchome:
@matthewoliver are you OK with current state so we can merge it?

xml/depl_maintenance.xml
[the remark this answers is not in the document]
matthewoliver: I'm fine with that too.

xml/depl_maintenance.xml
```diff
@@ -508,6 +507,289 @@
...
+       <command>psql2mysql migrate</command> without stopping the
+       database-related &ostack; services. This way, if the test migration
+       fails due to errors that weren't caught by the precheck procedure, you
+       can fix them with &ostack; services still running, thus minimizing the
+       required downtime. When you perform the actual migration, the data in
+       the target databases will be replaced with the latest one in the source databases.
+       </para>
+       <para>
+	After the test migration and before the actual migration, it is recommended to run the
+	<command>psql2mysql migrate</command> command to purge tables in the
```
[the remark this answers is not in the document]
dmpop: Fixed.

rsalevsky:
@dmpop Can we merge this into Cloud 8?

jsuchome:
When they are any future requests, we can create new PR...

dmpop:
All changes merged into maintenance/cloud_8

PR #559, opened 2018-10-11
fix RHEL installation instructions (bsc#1111503) (bsc#1111533)

csymons-suse:
add "/" to repository location
add missing dependency for RHEL Compute Nodes

### DIFF
--- a/xml/installation-installation-rhel-install_rhel_compute_node.xml
+++ b/xml/installation-installation-rhel-install_rhel_compute_node.xml
@@ -25,10 +25,10 @@
      Enable SMT mirroring of the external CentOS 7.3 repository to distribute
      RHEL-compatible packages to nodes in the cloud.
     </para>
-<screen>sudo smt-setup-custom-repos --name CentOS --description "CentOS 7.3" --productid 100682 \
-  --exturl https://download.opensuse.org/repositories/systemsmanagement:/Ardana:/8:/CentOS/CentOS_7.3
-sudo smt-repos -e CentOS
-sudo smt-sync
+<screen>&prompt.sudo;smt-setup-custom-repos --name CentOS --description "CentOS 7.3" --productid 100682 \
+  --exturl https://download.opensuse.org/repositories/systemsmanagement:/Ardana:/8:/CentOS/CentOS_7.3/
+&prompt.sudo;smt-repos -e CentOS
+&prompt.sudo;smt-sync
 sudo smt-mirror</screen>
    </step>
    <step>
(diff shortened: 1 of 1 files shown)


KeithMnemonic:
the libcacard is acutally on the install media, so if the repos are done correctly it will get pulled in automatically but only if you use yum. the notes here just list the 4 rpms and the rpm -U command.  WE should add more context here. something like libcard is needed and should be automtically pulled in when you use yum. If it is not found, you will need to add it manually,.....

csymons-suse:
The instructions preceding this section refer to the CentOS repository and YUM. So your comment must be related to some other process of getting to nova-compute installation. Is that correct? Does this matter to the documentation?

With the edits in this commit, the section reads:
```
The following RPM packages from that location have been tested successfully:
qemu-img-ev-2.6.0-28.el7.10.1.x86_64.rpm
qemu-kvm-common-ev-2.6.0-28.el7.10.1.x86_64.rpm
qemu-kvm-ev-2.6.0-28.el7.10.1.x86_64.rpm
qemu-kvm-tools-ev-2.6.0-28.el7.10.1.x86_64.rpm
libcacard-ev-2.3.0-31.el7_2.21.1.x86_64.rpm

Note that, in the case of migrating from a previous version of these packages, they should be installed using rpm -U FILENAME rather than rpm -i FILENAME.

All of the RPM packages should be installed together and not individually, as shown below:
tux > sudo rpm -Uvh qemu-img-ev-2.6.0-28.el7.10.1.x86_64.rpm \
qemu-kvm-common-ev-2.6.0-28.el7.10.1.x86_64.rpm \
qemu-kvm-ev-2.6.0-28.el7.10.1.x86_64.rpm \
qemu-kvm-tools-ev-2.6.0-28.el7.10.1.x86_64.rpm \
libcacard-ev-2.3.0-31.el7_2.21.1.x86_64.rpm
```
I just added:
```
The RPM package for libcacard-ev is required and should be automatically
pulled in when you use the Yum repo as described above. If this dependency
is not found, it must be installed manually with the qemu-kvm packages.
```
Does that get it?

KeithMnemonic:
Carl,

There are two ways to install packages on RHEL, One is using the "rpm" command and that is what we listed and we specifiy -U. If you are using rpm, it may not pull in libcacard automatically.  This is why you need to specify it with the other packages on one long rpm command. The other way is using the "yum" command This will pull in libcacard auotmatically if it finds it in one of its configured repos

csymons-suse:
One additional change:
```
The RPM package for libcacard-ev is required and should be automatically
pulled in when you use the Yum repo as described above <emphasis role="bold">and</emphasis> you install with the <command>Yum</command> command. If this dependency is not found, you will receive an error message reporting a failed dependency. In this case, the RPM package must be installed manually with the qemu-kvm packages.
```

spacefito:
+1

csymons-suse:
content moved to maintenance/cloud_8 in
https://github.com/SUSE-Cloud/doc-cloud/pull/585

PR #584, opened 2018-10-18
add CLM Admin UI (SCRD-4925)

csymons-suse:
this seems to fit Operations better than Installation where it is now. Opinions please. And where specifically should it go in whatever guide?


jeremy-moffitt:
Downloaded and built this, and it looks really sharp... though I agree with your instincts about it probably belonging in the Operations guide

PR #602, opened 2018-10-30
Add day2 ui content (SCRD-4468)

csymons-suse:


jeremy-moffitt:
I see `  <figure xml:id="clm-admin-add-server-3">`  which looks correct a glance because https://github.com/SUSE-Cloud/doc-cloud/blob/a825654eca32089da4170eda3d3808b6fa51f4de/images/src/png/clm_admin_add_server_3.png  is the correct image... but when I download and build the docs, I'm seeing one of the topology images in its place

jeremy-moffitt:
If it helps any, the image I'm seeing is clm_admin_topo_control_planes_0.png

PR #605, opened 2018-10-31
fix TLS disabled selection (bsc#1112812)

csymons-suse:


xml/security-tls_mysql.xml
```diff
@@ -104,7 +104,7 @@ mysql_gcomms_bind_tls: False</screen>
     </para>
 <screen># TLS disabled for cluster
 mysql_gcomms_bind_tls: "{{ host.bind['FND_MDB'].mysql_gcomms.tls }}"
-mysql_gcomms_bind_tls: False</screen>
+#mysql_gcomms_bind_tls: False</screen>
```
sknorr: Are you sure about this? The bug report seems to indicate that you should disable the other line, i.e. `"{{ host.bind[...`.

PR #612, opened 2018-11-06
add skip-unready-nodes (bsc#1068929)

csymons-suse:


xml/book_cloud_deploy.xml
```diff
@@ -38,6 +38,7 @@
   <xi:include href="depl_conf_admin_network.xml"/>
   <xi:include href="depl_conf_admin_crowbar.xml"/>
   <xi:include href="depl_crowbar_start.xml"/>
+  <xi:include href="depl_crowbar_extra-features.xml"/>
```
Itxaka: A better name for this section may be "Customizing the crowbar lifecycle manager"

xml/depl_crowbar_extra-features.xml
```diff
@@ -0,0 +1,82 @@
...
+   will print the log lines where nodes were skipped, the name of the node, and
+   the &barcl; which was being applied.
+  </para>
+  <screen>&prompt.user;grep "skipped until next chef run" /var/log/crowbar/production.log</screen>
+  <para>
+   You can also search the production log for skipped unready nodes.
+  </para>
+  <screen>&prompt.user;grep "skip_unready_nodes" /var/log/crowbar/production.log</screen>
+  <para>
+   After enabling/disabling the <literal>skip_unready_nodes</literal> feature
```
Itxaka: Can we wrap this section in one of those `Important` yellow boxes as to highlight it?

I think its important for the end user to be aware of that part.

xml/depl_crowbar_extra-features.xml
```diff
@@ -0,0 +1,82 @@
...
+  </para>
+  <para>
+   Skipped nodes are logged to the &crow; log
+   (<filename>/var/log/crowbar/production.log</filename>) where you can search
+   for the text <literal>skipped until next chef run</literal>. This
+   will print the log lines where nodes were skipped, the name of the node, and
+   the &barcl; which was being applied.
+  </para>
+  <screen>&prompt.user;grep "skipped until next chef run" /var/log/crowbar/production.log</screen>
+  <para>
```
Itxaka: I would remove this part and line 76 as there is not much info to obtain from that log, so it should not be communicated to the end user, its more for internal debugging.

xml/depl_crowbar_extra-features.xml
csymons-suse: Important tag around 
   After enabling/disabling the <literal>skip_unready_nodes</literal> feature
   or adding/removing roles, the &crow; framework service must be restarted
   (<command>systemctl restart crowbar</command>) in order to use the updated
   settings.

xml/depl_crowbar_extra-features.xml
csymons-suse: Are you suggesting to remove all of "Determining Which Nodes Were Skipped"?
Both production.log greps?

It seems like this is information that a user might want to know. I've taken out 'grep "skip_unready_nodes"', and left 'grep "skipped until next chef run"'

PR #625, opened 2018-11-13
restart Crowbar manually (bsc#1069434)

csymons-suse:
please check example for list_restarts, shouldn't keystone be flagged "false" if only apache2 is being restarted?

In the following example, the service to restart will be apache2.
<screen>crowbarctl services list_restarts
{
  "Node_IP_Example": {
    "alias": "controller1",
    "keystone": {
      "apache2": {
        "pacemaker_service": false,
        "timestamp": "2017-11-22 11:17:49 UTC"
      }
    }
  }
}

### DIFF
--- a/xml/depl_crowbar_extra-features.xml
+++ b/xml/depl_crowbar_extra-features.xml
@@ -25,7 +25,7 @@
    <emphasis role="bold">Enabling the Feature</emphasis>
   </para>
   <para>
-   In <filename>/opt/dell/crowbar_framework/config/crowbar.yml</filename> set
+   In <filename>/opt/dell/crowbar_framework/config/crowbar.yml</filename>, set
    the option <literal>skip_unready_nodes</literal> to <literal>true</literal>.
   </para>
 <screen>default: &amp;default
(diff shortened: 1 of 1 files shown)


Itxaka:
> please check example for list_restarts, shouldn't keystone be flagged "false" if only apache2 is being restarted?

Hey Carl,

This is expected. In crowbar the keystone service is just an apache2 using wsgi to launch the keystone python service, it doesnt has its own agent so its dependant on that.

The `false` in there signifies if this is a pacemaker service or a standalone service, usually pacemaker takes care of the service in HA environemnts, while on non-HA the services are launched by systemd.

Maybe we should clarify that the `pacemaker_service` item indicates if this service is launched by pacemaker or by systemd.

xml/depl_crowbar_extra-features.xml
```diff
@@ -105,4 +105,189 @@ skip_unready_nodes:
...
+}</screen>
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term>list_restarts</term>
+        <listitem>
+         <para>
+          Displays the list of pending restarts. The <literal>pending
+          restart</literal> flag indicates that a service tried to
+          restart due to the &chef; run but it did not due to the automatic
```
Itxaka: would add here:

`the pacemaker_service attribute indicates if this service is managed by pacemaker (usually in an HA environment) or its an standalone service managed by systemd(usually in on-HA environments)`

csymons-suse: Added. PDF attached.
[deployment_customizing.pdf](https://github.com/SUSE-Cloud/doc-cloud/files/2581404/deployment_customizing.pdf)

PR #637, opened 2018-11-17
change SLE 12 SP3 references to SLE 12 SP4 (SCRD-5138)

csymons-suse:


csymons-suse:
cloud 9 only (develop)

PR #636, opened 2018-11-16
SCRD-5619 Document migration w/OVSvApp + other small updates

cadenzajon:

PR #639, opened 2018-11-19
change HOS to ARDANA

csymons-suse:
https://github.com/SUSE-Cloud/doc-cloud/pull/636


cadenzajon:
This should definitely NOT be done. At this point in migration, the customer still has HOS 5.x installed, and so the HOS env vars and scripts need to be referenced.

PR #646, opened 2018-11-20
Manila requires restart after reconfiguration

spacefito:
If manila has already been installed and is being reconfigured, the service needs to be explicitly restarted.


spacefito:
addresses bsc#1116495

xml/installation-ardana-manila.xml
```diff
@@ -70,6 +70,10 @@
 &prompt.ardana;ansible-playbook -i hosts/verb_hosts ardana-gen-hosts-file.yml
 &prompt.ardana;ansible-playbook -i hosts/verb_hosts FND-CLU-reconfigure.yml
 &prompt.ardana;ansible-playbook -i hosts/verb_hosts ardana-reconfigure.yml</screen>
+    If &o_sharefs has already been installed but is being reconfigured run the following for
```
charleswang007: `&o_sharefs` needs to be `&o_sharefs;` I think

spacefito: done

PR #657, opened 2018-12-05
update netapp manila with DHSS=False & troubleshoot sections (bsc#1118201)

charleswang007:
This patch updates the doc with manila netapp installation steps with DHSS=False mode. It also adds a troubleshooting section that states what to do if compute does not have outside access to netapp servers.

### DIFF
--- a/xml/installation-ardana-manila.xml
+++ b/xml/installation-ardana-manila.xml
@@ -241,7 +241,8 @@ openstack-manila-share openstack-manila-scheduler
   </procedure>
   <para>
    Mode 2: The back-end is <literal>NetApp</literal>,
-   <literal>driver_handles_share_servers = True</literal> (DHSS is enabled).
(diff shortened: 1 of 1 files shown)


xml/installation-ardana-manila.xml
```diff
@@ -304,42 +305,17 @@ default_share_type = default1</screen>
...
-      NetAppException: Failed to create a route to 0.0.0.0/0 via gateway None:
-      Invalid value specified for <literal>gateway</literal> element within
-      <literal>net-routes-create</literal>
-     </para>
-    </listitem>
-   </varlistentry>
-  </variablelist>
+  <para>
+   if the compute nodes does not have access to manila back-end server, use the 
+   manila-share service on controller nodes instead. You can do so by either running
```
toabctl: What is the `manila back-end server` ?

charleswang007: It's the NetApp server in this case.

spacefito:
looks good

PR #664, opened 2018-12-10
DO NOT MERGE: Adding all images that were missing from DAPS build

asettle:


xml/admin-ironic.xml
```diff
@@ -2434,123 +2434,33 @@ nova flavor-key my-baremetal-flavor set capabilities:has_ssd="true"</screen>
                 </section>
               </section>
               <section>
-                <title>Deploy Process</title>
```
dmpop: So basically, **Deploy Process** consists of sections, where each section contains a single image. This means that if we remove the images, then we'll have section titles and nothing else. This might not be what we want. :-)

asettle: These are the "images" that are attempting to be rendered here: https://docs.openstack.org/ironic/pike/admin/drivers/ilo.html#id42

Now, the problem with these are they are HTML inserts that are read by RST and rendered. See here: https://github.com/openstack/ironic/blob/stable/pike/doc/source/admin/drivers/ilo.rst#netboot-with-glance-and-swift

These are not images we can "save" and process. At best I can get a bigger monitor, screen shot the image, and input it as a png file, but that's quite difficult in itself.

csymons-suse: > These are not images we can "save" and process. At best I can get a bigger
> monitor, screen shot the image, and input it as a png file, but that's 
> quite difficult in itself.

Some of the existing drawings were done with multiple screenshots (in higher resolution), stitched together in GIMP.

PR #667, opened 2018-12-12
Crowbar describe config snippets

toabctl:


toabctl:
Needs to be partly reverted because there is already a section about this topic.
Revert is https://github.com/SUSE-Cloud/doc-cloud/pull/669

PR #668, opened 2018-12-12
crowbar: Document the Salt barclamp

toabctl:
The Salt barclamp will be visible to Crowbar users so it needs some
basic documentation.


toabctl:
friendly ping - can we merge this?

PR #669, opened 2018-12-12
Drop troubleshot section about config snippets

toabctl:
This reverts b5bb2cd09923d019d9a8754c2e804974af548eaf because there is
already a section about configuration snippets. See
http://docserv.nue.suse.com/documents/SUSE_OpenStack_Cloud_9/suse-openstack-cloud-deployment/single-html/#cha.depl.ostack.configs

### DIFF
--- a/xml/depl_troubleshooting.xml
+++ b/xml/depl_troubleshooting.xml
@@ -834,52 +834,6 @@ EOF</screen>
      </answer>
     </qandaentry>
 
-    <qandaentry xml:id="sec.depl.trouble.faq.misc.change_openstack_config">
-     <question>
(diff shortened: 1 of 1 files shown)


fsundermeyer:
@csymons-suse: Some information from the deleted FAQ are missing in "cha.depl.ostack.configs" (for example how to restart the services). I also think that the FAQ itself is a valid one.
Therefor I would suggest:

1. Add the missing pieces from the deleted FAP to the chapter cha.depl.ostack.configs
2. Re-add the FAQ by keeping the question and making the answer a link to cha.depl.ostack.configs

PR #686, opened 2018-12-20
Fix SCRD-6491

dmpop:


xml/depl_maintenance.xml
```diff
@@ -407,6 +407,11 @@
        Upgrade Instructions</link>.
      </para>
     </listitem>
+    <listitem>
+      <para>
+	Run the command <command>nova-manage db archive_deleted_rows</command> to purge deleted instances from the database table. This can significanly reduce time required for the database migration procedure.
```
jsuchome: Please add also another commands (I know I did not mention them in the original request):
`cinder-manage db purge` (= Purge database entries that are marked as deleted)

dmpop: Done.

PR #695, opened 2018-12-27
Feature/clm replace servers

csymons-suse:
day 2 UI-replace server (SCRD-6684)
minor edits to day2UI-addserver


images/src/png/clm_replace_server.adoc
```diff
@@ -0,0 +1,186 @@
+Server Replacement
+------------------
+The process of replacing a server is initiated from the
+Server Summary. Replacing a will remove the existing server
```
jeremy-moffitt: one of us (probably me) seems to have had a typo here... "Replacing a will" should probably be "Replacing a server will"

csymons-suse: This is actually a different error than a typo. The file `clm_replace_server.adoc` should not be in the images/src/png directory. That is the source document that is not required anywhere in the repository. I've removed it.

The actual xml file (`xml/clm_replace_server.xml`) is correct...
  ```
The process of replacing a server is initiated from the Server Summary (see <xref
  linkend="admin_ui_servers"/>). Replacing a server will remove the existing
  server from the Cloud configuration and install the new server in its
  place. The rest of this process varies slightly depending on the type of
  server being replaced.
```

PR #709, opened 2019-01-17
Feature/upstream ostack docs convert

asettle:


asettle:
@dmpop that should be everything, but naturally it's going to fail against develop.

PR #722, opened 2019-01-29
SES 5.5 Integration-supersedes PR 718 (no bsc)

csymons-suse:
Supersedes PR 718


hemna:
Yes that looks good to me.

PR #731, opened 2019-02-01
SCRD-7115 Update RHEL 7.5 RPM locations and Ceph instructions

cadenzajon:


cadenzajon:
All occurrences of 7.3 have been fixed. Thanks for pointing that out.

PR #733, opened 2019-02-05
Fix/scrd 6661

dmpop:
Fix https://jira.suse.de/browse/SCRD-6661 (upgrade troubleshooting)


xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+    <qandaset>
+     <qandaentry>
+      <question>
+       <para>
+	Upgrade of the admin server has failed.
+       </para>
+      </question>
+      <answer>
+       <para>
+	Check for empty, broken, and not signed repositories in the &admserv;
```
jsuchome: "or" instead of "and"?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	upgrade log file <filename>/var/log/crowbar/admin-server-upgrade.log</filename>. Fix the
+	repository setup. Upgrade then remaining packages manually to &cloudos;
+	and &cloud; &productnumber; using the command <command>zypper dup</command>. Reboot the &admserv;.
+       </para>
+      </answer>
+     </qandaentry>
+     <qandaentry>
+      <question>
+       <para>
+	Upgrade step repeatedly fails due to timeout.
```
jsuchome: I would say "Some upgrade step", as there are more of them. But maybe my English is wrong in this case

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+        :set_network_agents_state: 300
+        :delete_pacemaker_resources: 600
+        :delete_cinder_services: 300
+        :delete_nova_services: 300
+        :wait_until_compute_started: 60
+        :reload_nova_services: 120
+        :online_migrations: 1800
+       </screen>
+       <para>
+	The following entries may require higher values:
```
jsuchome: It should be stated somewhere that the number means number of seconds

xml/depl_maintenance.xml
dmpop: Do you mean something like _a certain upgrade step_?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	Node upgrade has failed during live migration.
+       </para>
+      </question>
+      <answer>
+       <para>
+	The problem may occur when it is not possible to live
+	migrate certain VMs anywhere. It may be necessary to shut down or suspend
+	other VMs to make room for migration. Note that the Bash shell script
+	that starts the live migration for the &compnode; N is executed from
+	the &contrnode; M. Check the
```
jsuchome: We can probably leave those *N* and *M*'s as we're not using them in the text anyway.

jsuchome: And we should mention that both nodes are mentioned in the error message, something like, "Note that the Bash shell script that starts the live migration for the &compnode; is executed from the &contrnode; names of both nodes are mentioned in the error message that comes from 'crowbarctl upgrade status'"

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+       <para>
+	It is possible that live-migration of a certain VM takes too long. This
+	can happen if instances are very large or network connection between
+	compute hosts is slow or overloaded. If this case, try to raise the
+	global timeout in
+	<filename>/etc/crowbar/upgrade_timeouts.yml</filename>.  
+       </para>
+       <para>
+	We recommend to perform the live migration manually first. After it is
+	completed successfully, call the <command>crowbarctl upgrade</command> command.
```
jsuchome: "crowbarctl upgrade nodes"

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+       </para>
+       <para>
+	The following commands can be helpful for analyzing issues with live migrations:
+       </para>
+       <screen>
+	nova server-migration-list / server-migration-show
+        nova instance-action-list / instance-action
```
jsuchome: Please write those commands in full:
```
nova server-migration-list
nova server-migration-show
nova instance-action-list
nova instance-action```

File: xml/depl_maintenance.xml (hunk 2 above)
Comment (jsuchome) [reply xml/depl_maintenance.xml]: I don't know, is that better? I just mean that there is a number of upgrade steps, and any of them could time out

File: xml/depl_maintenance.xml [hunk 7]
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	Note that these commands require &ostack-bare; administrator privileges.
+       </para>
+       <para>
+	The following log files may contain useful information:
+       </para>
+       <itemizedlist>
+	<listitem>
+	 <para>
+	  <filename>/var/log/nova/nova-compute</filename> on the source and
+	  target &compnode;s.
```
jsuchome: Something like "migration source and target" so that it is clearer what we mean by "source and target". I still don't like it very much though.

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	 </para>
+	</listitem>
+	<listitem>
+	 <para>
+	  <command>/var/log/nova/*.log</command> (especially log files for the
```
jsuchome: Here you have "command" tag instead of "filename"

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	  conductor, scheduler and placement services) on the &contrnode;s.
+	 </para>
+	</listitem>
+       </itemizedlist>
+       <para>
+	It can happen that active instances and instances with heavy
+	loads cannot be live migrated in a reasonable time. In that case,  you
+	can abort a running live-migration using the <command>nova
```
jsuchome: "a" probably does not belong here?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	live-migration-about <replaceable>MIGRATION-ID</replaceable></command>
```
jsuchome: `nova live-migration-abort`

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	Node has failed during OS upgrade.
+       </para>
+      </question>
+      <answer>
+       <para>
+	Possible reasons include an incorrect repository setup or package
+	conflicts. Check the <filename>node-upgrade.log</filename> log file on the
```
jsuchome: `/var/log/crowbar/node-upgrade.log`

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	node. Check the repositories on node using the <command>zypper
```
jsuchome: problematic/affected node ?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	Node does not come up after reboot.
+       </para>
+      </question>
+      <answer>
+       <para>
+	It could happen that it takes the node too long to come up, which cause
+	a timeout. Check the node manually, and make sure it is online. Repeat then the step.
```
jsuchome: "Repeat the step" ?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	N number of nodes were provided to compute upgrade using
+	<command>crowbarctl upgrade nodes node_1,node_2,...,node_N</command>,
+	but less then N were actually upgraded.
+       </para>
+      </question>
+      <answer>
+       <para>
+	If the live migration cannot be performed for certain nodes due to a timeout,
+	&crow; upgrades only the nodes that it was able to
+	live-evacuate. Since some nodes have been upgraded, it is possible that
```
jsuchome: ... it was able to ￼live-evacuate **in time**

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	more resources will available for live-migration if you try to run this
+	step again. See also question on live migration failure.
```
jsuchome: BTW, is this block understandable? I understand it because I know what is it trying to describe (and because I wrote it), but I fear it's not written well.

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	Node has failed at the initial chef client run stage.
+       </para>
+      </question>
+      <answer>
+       <para>
+	An unsupported entry in the configuration file may prevent a service
+	from starting. This, in turn, causes the node to fail at the initial
+	chef client run stage. Checking the
+	<filename>/var/log/crowbar/crowbar_join/chef.*</filename> log files on
+	the node is a good starting point.
```
jsuchome: Do we have some troubleshooting for regular crowbar/chef deployments? If so, we could refer to it from this place (as it is less about upgrade itself)

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	Failure occurred when evacuating routers.
+       </para>
+      </question>
+      <answer>
+       <para>
+	Check the <filename>/var/log/crowbar/node-upgrade.log</filename> file on
+	the node that performs the router evacuation (it should be mentioned in
+	the error message). The ID of the router that failed to migrate (or the
+	affected network port) is logged to
+	<filename>/var/log/crowbar/node-upgrade.log</filename>). Use the
```
jsuchome: There's an extra ")" here

xml/depl_maintenance.xml
dmpop: I think _and_ is better, as you should check for all these symptoms and not just for one or two. Am I wrong?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+	&ostack-bare; CLI tools to check the state of the affected router and
+	its ports. Fix manually, if necessary. This can be done, for example,
+	by bringing the router or port up and down again. The following
+	commands can be useful for solving the issue: <command>openstack router
```
jsuchome: Could you format the commands so they is only one on a line? (Just like with nova-migrations above)

xml/depl_maintenance.xml
dmpop: OK. Then I think something simple _An upgrade step_ would work best.

xml/depl_maintenance.xml
jsuchome: Well, right. Your call. I guess my feeling for English is probably too much influenced by my Czech thinking :-)

xml/depl_maintenance.xml
dmpop: How about _on the compute nodes that the migration is perform from and to_? Not ideal, but maybe a bit more clear?

xml/depl_maintenance.xml
dmpop: I'm not sure where in the text _problematic/affected node_ should be.

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,295 @@ migrate</screen>
...
+      <question>
+       <para>
+	Upgrade status file is empty or corrupted. <command>crowbarctl upgrade status</command> reports invalid information.
+       </para>
+      </question>
+      <answer>
+       <para>
+	This can happen if the &admserv; runs out of disk space or &crow;
+	service is stopped abruptly during a status write operation. Contact
+	support for assistance on how to recover the status file.
```
jsuchome: I'm not sure if this answer is helpful at all. I'm inclined to delete the paragraph completely. @rhafer @skazi0 ideas?

xml/depl_maintenance.xml
dmpop: I understand it, and I'm not even an expert. :-)

xml/depl_maintenance.xml
jsuchome: Sounds better, yes

xml/depl_maintenance.xml
jsuchome: Check the <filename>node-upgrade.log</filename> log file on the _node_

Just wanted to make sure we know what node should be checked

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+       </para>
+      </answer>
+     </qandaentry>
+     <qandaentry>
+      <question>
+       <para>
+	After performing <command>crowbarctl upgrade nodes
+	controllers</command>, some non-controller nodes weren't upgraded.
```
jsuchome: Here really is "were". The point was that user might be surprised to see more nodes upgraded that was requested

xml/depl_maintenance.xml
skazi0: @jsuchome we don't want to write full "how to make a status file by hand" guide, do we? The idea behind this hint was to make people aware that this file could be sometimes re-created if lost. If you think that this doesn't bring much benefit then just drop this sentence.

jsuchome: > we don't want to write full "how to make a status file by hand" guide, do we? 

Yeah, that sounds too dangerous. People could get easily into unsupported scenario.

I'm honestly not sure. I guess when they find out that their upgrade is broken, they know to contact us anyway.

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
     the selected nodes.
    </para>
   </sect2>
+  <sect2 xml:id="sec.depl.maintenance.upgrade.troubleshooting">
+   <title>Common Upgrade Problems and Possible Solutions</title>
```
asettle: "possible" ?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	Upgrade of the admin server has failed.
+       </para>
+      </question>
+      <answer>
+       <para>
+	Check for empty, broken, and not signed repositories in the &admserv;
+	upgrade log file <filename>/var/log/crowbar/admin-server-upgrade.log</filename>. Fix the
```
asettle: "... in the &admserv; and upgrade log file..

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	repository setup. Upgrade then remaining packages manually to &cloudos;
```
asettle: What does "fix the repository setup" mean?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+       </para>
+	<itemizedlist>
+	 <listitem>
+	  <para>
+	   <literal>upgrade_os</literal> Time allowed for upgrading all packages of one node.
+	  </para>
+	 </listitem>
+	 <listitem>
+	  <para>
+	   <literal>chef_upgraded</literal> Time allowed for initial crowbar_join and chef-client run on a node that has been upgraded and rebooted.
```
asettle: Should crowbar_join and chef-client be marked up?

asettle: Also does this sentence end here or is it running off the page?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	<listitem>
+	 <para>
+	  <filename>/var/log/nova/*.log</filename> (especially log files for the
+	  conductor, scheduler and placement services) on the &contrnode;s.
+	 </para>
+	</listitem>
+       </itemizedlist>
+       <para>
+	It can happen that active instances and instances with heavy
+	loads cannot be live migrated in a reasonable time. In that case,  you
```
asettle: extra space between the comma and "you"

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	Node does not come up after reboot.
+       </para>
+      </question>
+      <answer>
+       <para>
+	It could happen that it takes the node too long to come up, which cause
+	a timeout. Check the node manually, make sure it is online, and repeat the step.
```
asettle: "In some cases,  a node can take too long to reboot causing a timeout. We recommend to check the node manually..."

xml/depl_maintenance.xml
asettle: Sorry folks, I don't understand it. What does "...upgrades only the nodes that it was able to live-evacuate in the specified time." mean?

asettle: "See also question on live migration failure." -> "See the question on live migration failure for more information" (and then link to that question if possible)

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	Node has failed at the initial chef client run stage.
+       </para>
+      </question>
+      <answer>
+       <para>
+	An unsupported entry in the configuration file may prevent a service
+	from starting. This, in turn, causes the node to fail at the initial
```
asettle: Remove "in turn" - unnecessary verbiage.

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+     </qandaentry>
+     <qandaentry>
+      <question>
+       <para>
+	I need to change &ostack-bare; configuration during the upgrade but I cannot access &crow;.
+       </para>
+      </question>
+      <answer>
+       <para>
+	&crow; &wi; is accessible only when upgrade is completed or
```
asettle: only when an upgrade is completed*

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	when it is postponed. Postponing the upgrade can be done only after
+	upgrading all &contrnode;s using the <command>crowbarctl upgrade nodes
+	postpone</command> command. You can then access &crow; and
+	even save some modification. It is then necessary to resume the upgrade
```
asettle: "You can then access &crow; and save your modifications. Before you can continue with the upgrade of rest of the nodes, resume the upgrade using the ...."

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+      <question>
+       <para>
+	Failure occurred when evacuating routers.
+       </para>
+      </question>
+      <answer>
+       <para>
+	Check the <filename>/var/log/crowbar/node-upgrade.log</filename> file on
+	the node that performs the router evacuation (it should be mentioned in
+	the error message). The ID of the router that failed to migrate (or the
```
asettle: And if it isn't mentioned on the error message?

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	affected network port) is logged to
+	<filename>/var/log/crowbar/node-upgrade.log</filename>. Use the
+	&ostack-bare; CLI tools to check the state of the affected router and
+	its ports. Fix manually, if necessary. This can be done, for example,
```
asettle: Remove "for example"

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	<para>
+	 Resume the upgrade by running the failed upgrade step
+	again to continue with the router migration.
+       </para>
+      </answer>
+     </qandaentry>
+     <qandaentry>
+      <question>
+       <para>
+	Upgrade status file is empty or corrupted. <command>crowbarctl upgrade status</command> reports invalid information.
```
asettle: wrap the text here please

xml/depl_maintenance.xml
asettle: Just include a "how to contact support" note. Seems obvious - but people want the information *there*

xml/depl_maintenance.xml
asettle: "Some nodes were not upgraded after performing <command>crowbarctl upgrade nodes controllers</command>." (keep the language simple like the above examples)

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+     <qandaentry>
+      <question>
+       <para>
+	After performing <command>crowbarctl upgrade nodes
+	controllers</command>, some non-controller nodes weren't upgraded.
+       </para>
+      </question>
+      <answer>
+       <para>
+	In current upgrade implementation, &ostack-bare; nodes are divided into
```
asettle: In the current upgrade implementation*

xml/depl_maintenance.xml
```diff
@@ -1503,6 +1503,304 @@ migrate</screen>
...
+	&compnode;s and other nodes. The <command>crowbarctl upgrade nodes
+	controllers</command> command starts upgrade of all nodes that do not
+	host compute services are upgraded along with the controllers.
```
asettle: "...command starts the upgrade of all the nodes that do not host compute services. This includes the controllers." (minimize sentence repetition)

xml/depl_maintenance.xml
asettle: I would recommend this is retitled to "Troubleshooting upgrades"

xml/depl_maintenance.xml
jsuchome: We don't really know what went wrong, that's why the advice is generic. It's possible repositories are empty for example.
User should know what "fix the repo setup" means as he/she was supposed to prepare those repos before

xml/depl_maintenance.xml
dmpop: > Sorry folks, I don't understand it. What does "...upgrades only the nodes that it was able to live-evacuate in the specified time." mean?

It means that it can only upgrade the nodes that can be evacuated before hitting the specified timeout. Is that correct @jsuchome ?

xml/depl_maintenance.xml
dmpop: If the solution is to contact support, then it's not a solution at all. In which case, I vote for removing this part altogether.

xml/depl_maintenance.xml
dmpop: Good call. How about _Troubleshooting Upgrade Issues_? (You don't troubleshoot upgrades, no?)

xml/depl_maintenance.xml
jsuchome: When more nodes are provided for parallel upgrade (in the way of `crowbarctl upgrade nodes node1,node2,node3`) crowbar first tries to live-evacuate each node one by one (so _not_ in parallel). After that it means they do not have any VM's running so it can upgrade all of them. 

live-evacuation of a node is live-migration of each VM from such node

xml/depl_maintenance.xml
jsuchome: I think it is. So we can change the wording probably.

xml/depl_maintenance.xml
dmpop: I think the original version is correct. Basically it means that you have to check the upgrade log file on the admin server. @jsuchome Correct?

jsuchome: Yes, the /var/log/crowbar/admin-server-upgrade.log log file is located on admin server

xml/depl_maintenance.xml
jsuchome: > It means that it can only upgrade the nodes that can be evacuated before hitting the specified timeout. Is that 
> correct @jsuchome ?

Right. The main point of this issue is that it could be a subset of the nodes that user wanted.

xml/depl_maintenance.xml
jsuchome: I'd say drop it

xml/depl_maintenance.xml
asettle: Good point - suggestion approved! :D

xml/depl_maintenance.xml
```diff
@@ -1625,8 +1627,8 @@ migrate</screen>
        <itemizedlist>
 	<listitem>
 	 <para>
-	  <filename>/var/log/nova/nova-compute</filename> on the source and
-	  target &compnode;s.
+	  <filename>/var/log/nova/nova-compute</filename> on the &compnode;s
+	  that the migration is performed from and to.
```
asettle: "... on the source and origin &compnode;s."

xml/depl_maintenance.xml
asettle: Ah I see it now :)

xml/depl_maintenance.xml
asettle: Hmm should we make this clear? Or do we assume it'll be clear enough?

xml/depl_maintenance.xml
asettle: That'd be wise. I think all we need to do is remove the comment in the ( )

xml/depl_maintenance.xml
dmpop: _the &compnode;s that the migration is performed from and to_ is admittedly a bit tortured, but it conveys the intended meaning better, methinks.

xml/depl_maintenance.xml
jsuchome: No, this is not right, please see my original comment.

PR #735, opened 2019-02-06
Removing male specific pronouns in favor of neutral terms

asettle:


asettle:
There were a few more within the repo but those files are deleted in this PR https://github.com/SUSE-Cloud/doc-cloud/pull/734

PR #751, opened 2019-02-12
Upstream guides: Updates from Colleen

asettle:
Reference: https://etherpad.nue.suse.com/p/soc8-concerns


xml/user-glance.xml
```diff
@@ -835,7 +835,7 @@ managed at a namespace level.
                     mapping in the following format:</para>
             <screen>{'images': [
   {'uri': 'http://glance.openstack.example.org/v1/images/71c675ab-d94f-49cd-a114-e12490b328d9',
-   'name': 'Ubuntu 10.04 Plain',
+   'name': 'SLE 15',
```
cmurphy: Not positive but it might be better to call it SLES 15 or SUSE Linux Enterprise Server 15

cmurphy: In our engineering cloud we call it SLES15

asettle: Ah - I shouldn't have been so literal. Sorry.

asettle: @dmpop do we have official style guide naming conventions we need to stick to here?

dmpop: We have entities for both SLE and SUSE Linux Enterprise Server. But I'm not sure whether it makes sense to use them here. Just use SLES 15.

asettle:
Note: https://github.com/SUSE-Cloud/doc-cloud/blob/ee1a3cccc9a566d75a2ad406c1f374c01d38289b/xml/admin-neutron-config-dhcp-ha.xml

PR #757, opened 2019-02-13
Rename “Installing with CLM” to “Deployment Guide using CLM”

asettle:
This PR changes the entity declaration for "soclcminstall" so that it no longer says "Installing with &lcm;" but rather, "Deployment Guide using &lcm;". This is because we are updating the name of the Deployment guide and will infer a more permanent change than simply updating the book file.


asettle:
This change matches the one made for Crowbar: https://github.com/SUSE-Cloud/doc-cloud/pull/706/

PR #759, opened 2019-02-13
Create script to build upstream OpenStack docs w/SUSE theme (SCRD-6893)

jeremy-moffitt:
WIP version of the script to download, brand, and produce HTML docs for each of the upstream OpenStack projects that are used by SUSE OpenStack Cloud. 

This is not (yet at least) built into any sort of container, we'll have to decide if thats worthwhile later.
Also, I'm still working out issues building Keystone docs, and Alexandra has a PR up to fix an issue with the Swift docs, so those are excluded for now.


scripts/gen_docs.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+# Author: Thomas Bechtold <tbechtold@suse.com>
+# License: Apache-2.0
+
+# List of project for running the doc generation
+: ${PROJECTS:="nova horizon cinder glance neutron manila aodh barbican ceilometer designate heat ironic magnum murano neutron-fwaas neutron-lbaas sahara"}
+# swift docs don't build successfully on stable/rocky
```
asettle: Dependent on: https://review.openstack.org/636595

scripts/gen_docs.sh
```diff
@@ -0,0 +1,68 @@
...
+# keystone docs dont build successfully with the script, requires python3, which should be whats being used
```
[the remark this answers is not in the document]
jeremy-moffitt: I thought it was, I'm still debugging this one, hopefully will have an answer today :)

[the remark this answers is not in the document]
jeremy-moffitt: turns out I was missing openldap2-devel , going to add keystone back into the list and update the keystone comment to suggest that a failure to build Keystone docs successfully may mean the openldap2-devel package is missing.

csymons-suse:
We no longer support
aodh

Services we say we support are missing:
horizon
monasca
monasca-ceilometer
neutron-VPNaaS
nova
octavia
swift

Services here that we _don't_ say we support:
murano
sahara

This list should be checked by Product Management

jeremy-moffitt:
> We no longer support
> aodh
> 
> Services we say we support are missing:
> horizon
> monasca
> monasca-ceilometer
> neutron-VPNaaS
> nova
> octavia
> swift
> 
> Services here that we _don't_ say we support:
> murano
> sahara
> 
> This list should be checked by Product Management

I based the list off of what we're currently packaging with Cloud 9 at https://build.suse.de/project/show/Devel:Cloud:9:Staging  (filter for "-doc") , I'm happy to adjust it though. 

Swift is MIA because of the build issue
Nova and Horizon are in the list (I need to fix the ordering, its currently "big name services" then "alphabetical list of the rest" which is confusing).

I'll email PM and some of the packaging folks to understand the mismatch between supported services and whats being packaged.

jeremy-moffitt:
I was indeed missing neutron-vpnaas , good catch! Asking about the others

scripts/gen_docs.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+# Author: Thomas Bechtold <tbechtold@suse.com>
```
GarySmith: You may want to adjust/add to this authorship depending on your relative contribution

scripts/gen_docs.sh
```diff
@@ -0,0 +1,68 @@
...
+        git checkout doc/source/conf.py
+        echo "### Updating conf.py"
+        sed -i "s/html_theme = 'openstackdocs'/html_theme = 'suse_sphinx_theme'/g" doc/source/conf.py
+        if [ -f doc/requirements.txt ]; then
+            echo "### Updating doc/requirements.txt"
+            git checkout doc/requirements.txt
+            grep -q -x -F 'suse_sphinx_theme' doc/requirements.txt || echo '-e git+https://github.com/SUSE/suse-sphinx-theme.git@master#egg=suse_sphinx_theme' >> doc/requirements.txt
+        else
+            if [ -f test-requirements.txt ]; then
+               echo "### Updating requirements.txt"
```
GarySmith: Nit: this should say "Updating test-requirements.txt"

scripts/gen_docs.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Author: Thomas Bechtold <tbechtold@suse.com>
+# License: Apache-2.0
+
+# List of project for running the doc generation
+: ${PROJECTS:="nova horizon cinder glance neutron manila aodh barbican ceilometer designate heat ironic magnum murano neutron-fwaas neutron-lbaas sahara"}
+# swift docs don't build successfully on stable/rocky
+# keystone docs dont build successfully with the script, requires python3, which should be whats being used
+# directory where the OpenStack git repositories are in
+: ${OPENSTACK_REPO_DIR:=${PWD}/devel/openstack}
```
GarySmith: I don't see a reason to create two levels of directory structure above the additional ones that will be created by cloning the various projects. But if you decide to keep these, then the default output dir should also be within this directory structure as well (e.g. ${PWD}/devel/openstack-docs-html) to avoid creating **two** new deeply nested directory structures.

jeremy-moffitt:
addressed most of the issues, I *sometimes* see the following error building Ironic (its not 100% of the time, but seems to happen most 'clean' runs but never again after):
```Warning, treated as error:
ironic/drivers/modules/iscsi_deploy.py:docstring of
 ironic.drivers.modules.iscsi_deploy.AgentDeployMixin.continue_deploy:1:
duplicate object description of AgentDeployMixin.continue_deploy, 
other instance in ironic/doc/source/contributor/api/ironic.drivers.modules.agent.rst, 
use :noindex: for one of them
ERROR: InvocationError: 
'ironic/.tox/docs/bin/sphinx-build -W -b html doc/source doc/build/html'
```

scripts/gen_docs.sh
jeremy-moffitt: added myself, don't really want to remove an author attribution, though Tom might want me to if he sees how I've mangled his script ;-)

scripts/gen_docs.sh
jeremy-moffitt: put them both under devel for now

GarySmith: Thanks

scripts/gen_docs.sh
jeremy-moffitt: fixed in latest version

scripts/gen_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+
+set -e
+
+# Author: Thomas Bechtold <tbechtold@suse.com>
+# Author: Jeremy Moffitt <jmoffitt@suse.com>
+# License: Apache-2.0
+
+# List of project for running the doc generation
+#: ${PROJECTS:="aodh barbican ceilometer cinder designate glance heat horizon keystone magnum manila murano neutron neutron-fwaas neutron-lbaas neutron-vpnaas octavia nova sahara"}
+: ${PROJECTS:="keystone manila"}
```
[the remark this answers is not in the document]
aspiers: Yeah, e.g.

    source ~/.config/suse-openstack-docs

scripts/gen_docs.sh
aspiers: `devel` is a bit too vague for my tastes.  How about `git` or `upstream` or `upstream-git`?

sknorr:
Thanks for your efforts!

I have a few conceptual (perhaps ignorant, sorry) questions:

1. How is the whole workflow going to look?

2. In particular, how are you making sure all deps are available and working? Iirc, that was one of the bigger problems for us with the previous iteration of Thomas's script. Missing deps is also what I just experienced trying this out -- it only worked after manually installing python3-tox, python3-devel, libldapcpp-devel & libopenssl-1_1-devel. And that is with just two of the OpenStack modules enabled -- I believe when we tried the earlier iteration of the script that cloned more modules, there were a lot more dependencies and some of them were either not available on openSUSE not available in the right version. Has that situation significantly changed? Is there a way to get around installing all those dependencies when you're just interested in creating docs.

3. Would it be a make sense to have the script create a venv before building? (Might become irrelevant if this is going to run in a container anyway.)

4. How is this script going to produce the expected output documents? Right now, this script builds each individual module guide into its own little folder. As far as I know, PM wants exactly two guides to come out at the end: an Admin Guide and a User Guide.
During previous conversion efforts, that was one of the most troublesome parts -- it is not easy to pull a guide into two, throw away some of the content, keep cross-references and still come up with something worth shipping.
To be clear: The Admin/User Guide split is a requirement that came from PM; the doc team would be happy to just ship the upstream guides.

5. Did you consider the necessity for downstream patches? For example, the upstream guides may describe features not available in SOC or OSs not supported by SOC. This might create unrealistic support expectations.

scripts/gen_docs.sh
jeremy-moffitt: changed to upstream

scripts/gen_docs.sh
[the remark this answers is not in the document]
jeremy-moffitt: good idea, moved these out to a .conf file, for the moment I put it in the scripts folder because I wanted to avoid creating another folder just for the one conf file after creating a scripts folder just for the scripts file... but I'm open to revisiting that... also realized that after splitting them out the script doesnt run correctly if its run from another directory since it looks for the conf file in the working folder... might need to revisit that

scripts/gen_docs.sh
```diff
@@ -0,0 +1,86 @@
...
+            cd ${OPENSTACK_REPO_DIR}
+            echo "### Cloning git repo"
+            git clone https://git.openstack.org/openstack/${project}
+        fi
+        cd ${OPENSTACK_REPO_DIR}/${project}
+        git checkout ${OPENSTACK_REPO_BRANCH}
+        echo "### Clear out any changes that may have come from previous failed runs"
+        git checkout doc/source/conf.py
+        echo "### Updating conf.py"
+        sed -i "s/html_theme = 'openstackdocs'/html_theme = 'suse_sphinx_theme'/g" doc/source/conf.py
```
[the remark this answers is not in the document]
jeremy-moffitt: done! thanks

jeremy-moffitt:
> Thanks for your efforts!
> 
> I have a few conceptual (perhaps ignorant, sorry) questions:
> 
> 1. How is the whole workflow going to look?

presumably the nightly build would kick this off as part of whatever process kicks off the builds now, but I confess to a complete lack of knowledge how that works

> 2. In particular, how are you making sure all deps are available and working? Iirc, that was one of the bigger problems for us with the previous iteration of Thomas's script. Missing deps is also what I just experienced trying this out -- it only worked after manually installing python3-tox, python3-devel, libldapcpp-devel & libopenssl-1_1-devel. And that is with just two of the OpenStack modules enabled -- I believe when we tried the earlier iteration of the script that cloned more modules, there were a lot more dependencies and some of them were either not available on openSUSE not available in the right version. Has that situation significantly changed? Is there a way to get around installing all those dependencies when you're just interested in creating docs.
> 3. Would it be a make sense to have the script create a venv before building? (Might become irrelevant if this is going to run in a container anyway.)

I think the best solution for 2&3 is to run this in a container. I wanted to get the script mostly working before delving into what would be involved in the container setup

> 4. How is this script going to produce the expected output documents? Right now, this script builds each individual module guide into its own little folder. As far as I know, PM wants exactly two guides to come out at the end: an Admin Guide and a User Guide.
>    During previous conversion efforts, that was one of the most troublesome parts -- it is not easy to pull a guide into two, throw away some of the content, keep cross-references and still come up with something worth shipping.
>    To be clear: The Admin/User Guide split is a requirement that came from PM; the doc team would be happy to just ship the upstream guides.

This was a big question that I totally overlooked before it was mentioned this morning. I'm not sure that theres a good way collate the output into a single guide... My (mistaken it seems) understanding was that the branding aspect was what was causing a lot of the manual effort and that if we could just build the upstream docs automatically then we wouldnt need 6-8 months after release to get the guides out. This warrants some discussion around feasibility.  

> 5. Did you consider the necessity for downstream patches? For example, the upstream guides may describe features not available in SOC or OSs not supported by SOC. This might create unrealistic support expectations.

We talked about trying to push SUSE specific doc pieces upstream where possible. There are platform flags that we may need to add to the build in support of this. Beyond that, it should be possible to create a patches folder with subfolders for each project and apply those patches before building. Whether its worth updating the script to do that probably depends on how effectively we can solve question 4.

scripts/gen_docs.sh
asettle: Update: Patch has been merged.

jeremy-moffitt:
based on this mornings meeting, the next iteration will focus on creating admin/user guide specific folders, followed by creating a template to act as an index for each of those (an admin guide index, and a user guide index)

jeremy-moffitt:
latest version adds the following:
1. separates out the admin/user content into their own folders
2. removes links to content outside of the admin guide (for admin docs) and outside of the user guide (for user docs)

script does not yet create a parent "table of contents" type page... still on the TODO list

GarySmith:
> I was indeed missing neutron-vpnaas , good catch! Asking about the others

It turns out you weren't really missing it. VPNaaS is covered in the [Miscellaneous section of the Neutron guide](https://docs.openstack.org/neutron/latest/admin/misc.html), and the doc that is generated from the neutron-vpnaas is basically an html doc consisting of a link back to the neutron doc. I think it should be omitted.

GarySmith:
> @GarySmith I think a file is missing maybe:
> 
> ```
> ##### Table of Contents
> sed: can't read /admin.html: No such file or directory
> ```

Somehow I overlooked putting the templates into my git commit. Some styling needs to be done on them before they could be released in their current barebones form.

jeremy-moffitt:
looks like this change will need to be rebased (the Travis CI job is failing due to some files that have been removed that are independent of this change)

GarySmith:
@jeremy-moffitt Check out this latest version when you get a chance. There is one known problem: the ironic documents are not styled correctly due to the fact that the regex for updating conf.py is not quite smart enough; I'll fix that tomorrow.
Note that the top-level generated docs are in `scripts/build/admin/html/index.html` and `scripts/build/user/html/index.html`

GarySmith:
The docker container in the last commit addresses SCRD-6893.

scripts/gen_docs.sh
```diff
@@ -0,0 +1,233 @@
...
+# Author: Gary Smith <gary.smith@suse.com>
+# License: Apache-2.0
+
+# extra tox options (eg. -r)
+: ${TOX_OPTS:=""}
+
+# Create entries for the table of contents doc that will be generated
+declare -A summary
+summary[barbican]='Key Manager service (barbican)'
+summary[ceilometer]='Telemtry Data Collection service (ceilometer)'
```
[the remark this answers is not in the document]
GarySmith: Good catch.

scripts/gen_docs.sh
```diff
@@ -0,0 +1,233 @@
...
+# Author: Thomas Bechtold <tbechtold@suse.com>
+# Author: Jeremy Moffitt <jmoffitt@suse.com>
+# Author: Gary Smith <gary.smith@suse.com>
+# License: Apache-2.0
+
+# extra tox options (eg. -r)
+: ${TOX_OPTS:=""}
+
+# Create entries for the table of contents doc that will be generated
+declare -A summary
```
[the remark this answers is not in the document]
GarySmith: It defines the titles only. The order is dictated by the PROJECTS environment variable, which may be read from the config file

GarySmith:
> Thanks for the update -- I tried the script again. Unfortunately, there does not seem to be a README for how to properly use it.
> 
> * running gen_docs.sh directly on the local machine:
>   
>   * seems to work, but takes forever (at least an hour) -- if you could speed this up in any way (is there no option for tox to run less stuff? I can't imagine building just documentation can take this long)
>   * after some time, the script quit on me because I was missing some dependency (sort of my fault) -- in any cause, I lost patience & did not see the output at the end
>     **=> need to try again**
> * running in docker:
>   
>   * I tried `docker build .` and then `docker run IMAGE_HASH`
>   * when building, I noticed that it was copying a lot of stuff (5 GB) into the image, presumably those were all the OpenStack repos I had just cloned while running the script locally -- not sure if that is a useful behavior
>   * `docker run` failed with `"exec: \"/doc-cloud/scripts/gen_docs.sh\": stat /doc-cloud/scripts/gen_docs.sh: no such file or directory"`
>     **=> no dice here either, this time I don't think it was my mistake but correct me on that**

There is a README in the scripts dir, and it shows the two commands that are needed to run the container via docker-compose, which also maps /doc-cloud properly

sknorr:
Ok, so just to recap (I think I said many of these things on the call yesterday) already:

* *regular HTML:* 
  * looks pretty reasonable and imo works well enough
  * a number of navigational elements still lead into unsupported territory. I think there is no way around that however. Maybe there is a way to add some disclaimer to the pages that are not supported -- if you can single them out..?
  * I am a bit worried about what will happen when Google eventually finds the unsupported pages
* *single-HTML:* This one seems to be more critical:
  * these are actually multiple HTMLs, each containing the entire guide for one of the OpenStack components (including all the stuff we don't want afaict)
  * merging all the component single-HTMLs into one big one is a hard problem:
    * doing this within the build process: would mean that all components have to be built together somehow
    * after the build: merging all the generated single-HTMLs via some XML tool might probably be ~easier on the face of it, however, it would also break all internal links in the document.
  * there is all the content in there that we don't want and there is likely no reasonable way around that.

* PDF (TR had already agreed on yesterday's call to deprioing this):
  * This would probably be relatively easy to generate from a real single-HTML, e.g. via wkhtmltopdf or just a regular browser, but given that single-HTML is such a big problem, it is not really attainable via this way either.
  * Generating PDFs via Sphinx would run into all the same content issues that already exist in the single-HTML build. In addition, you'd also have to work on branding again. I think Sphinx is using Latex for PDF output, so that endeavor is likely going to be decidedly un-fun.

=> I would reconsider whether trying to build single-HTML and PDF is really an option as opposed to just a waste of time.

jeremy-moffitt:
may need rebased 
```
The following DC file(s) from the repository are not valid:
* The "MAIN" file referenced in DC-suse-openstack-cloud-socmmsoperator does not exist.
* The "MAIN" file referenced in DC-suse-openstack-cloud-socmoverview does not exist.
```

dirkmueller:
I've merged this into https://github.com/SUSE-Cloud/doc-cloud-upstream/commits/rocky . lets keep this separate in a new git.

PR #763, opened 2019-02-14
add Cassandra recovery steps (bsc#1123607,SCRD-7656)

csymons-suse:


csymons-suse:
PDF preview attached to PR 762

PR #772, opened 2019-02-19
added doc for pci-passthrough

sjamgade:


xml/pci-passthrough
```diff
@@ -0,0 +1,83 @@
...
+for eg: it is common to have the *nouveau* driver installed while the node was installed.
+this driver is a graphics driver for Nvidia based GPU. And it needs to be blacklisted
+
+`echo 'blacklist nouveau' >> /etc/modprobe.d/nouveau-default.conf`
+
+the name of the file is not really important want matters is the location of file and contents
+similarly other drivers can be blacklisted (including Nvidia drivers on the node)
+
+##### Check 2:
+`iommu_groups` should be enanbled on the host, to check if iommu is enabled
```
vuntz: typo: enanbled

xml/pci-passthrough
```diff
@@ -0,0 +1,83 @@
...
+this shows that the devices are available for passthrough
+
+NOTE: when deailing with PCI passthrough a only whole iommu group can be passed and no part of it, here it is group 20.
+
+
+### Part 2
+
+Preparing Nova and Glance for passthrough:
+
+https://docs.openstack.org/nova/rocky/admin/pci-passthrough.html
```
vuntz: Is there anything to be done to make sure that a device is only used by one specific instance (which I hear would be required for GPU)? I don't see any doc about this.

sjamgade: once the compute node is configured correctly with `alias` and `pcipassthough_whitelist` it reports count and stats based on whitelist to scheduler, and scheduler does the book-keeping of what is used and what is available. 

that link mentions configuring nova-scheduler
As soon as nova-scheduler uses `PCIPassthoughFilter` it starts filtering nodes which have the required pci device while creating an instance. Since it also keeps track of what is assigned the filter rejects if no nodes with the required pci device are available or all available are already assigned

vuntz: Ah, I missed the link in the link. Which I think may mean that others will miss it... Is there any reason we don't use `PCIPassthoughFilter` by default?

(my concern really is that it's easy to make a mistake while following this doc -- but that's not a blocker for the pull request, it's more about how we deliver the feature)

sjamgade: yes, its possible to miss the second link, I will add a note about configuring  both.: scheduler and compute

There are is no reason for that filter to be missed, but there are other filters like CoreFilter and RamFilter which are also required for `*_raio` settings to have an effect.

I will create a PR for PCI passthrough filter

csymons-suse:
superseded by
https://github.com/SUSE-Cloud/doc-cloud/pull/775

PR #778, opened 2019-02-21
adoc content converted to DocBook XML

csymons-suse:
supersedes PR 772
additional information and PDF in
PR 775

### DIFF
--- a/xml/networking-pcipt_sriov_config.xml
+++ b/xml/networking-pcipt_sriov_config.xml
@@ -115,6 +115,115 @@
    process the proper VLAN ID, which therefore has to be known in advance
   </para>
  </section>
+ <section xml:id="_pci_passthrough">
+  <title>Leveraging PCI Passthrough</title>
+  <para>
+   Two parts are necessary to leverage PCI passthrough on a &cloud; 8
+   &compnode;: preparing the &compnode;, preparing &o_comp; and &o_img;.
+  </para>
+  <orderedlist>
+   <listitem>
+    <para>
+     <emphasis role="bold">Preparing the &compnode;</emphasis>
+    </para>
+    <procedure>
+     <step>
+      <para>
+       There should be no kernel drivers or binaries with direct access to the
+       PCI device. If there are kernel modules, they should be blacklisted.
+      </para>
+      <para>
+       For example, it is common to have a <literal>nouveau</literal> driver
+       from when the node was installed. This driver is a graphics driver for
+       Nvidia-based GPUs. It must be blacklisted as shown in this example.
+      </para>
+<screen>&prompt.ardana;echo 'blacklist nouveau' &gt;&gt; /etc/modprobe.d/nouveau-default.conf</screen>
+      <para>
+       The file location and its contents are important; the name of the file
+       is your choice. Other drivers can be blacklisted in the same manner,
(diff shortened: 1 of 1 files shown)


xml/networking-pcipt_sriov_config.xml
```diff
@@ -115,6 +115,115 @@
...
+      </para>
+      <para>
+       For example, it is common to have a <literal>nouveau</literal> driver
+       from when the node was installed. This driver is a graphics driver for
+       Nvidia-based GPUs. It must be blacklisted as shown in this example.
+      </para>
+<screen>&prompt.ardana;echo 'blacklist nouveau' &gt;&gt; /etc/modprobe.d/nouveau-default.conf</screen>
+      <para>
+       The file location and its contents are important; the filename is your
+       choice. Other drivers can be blacklisted in the same manner, possibly
```
asettle: the filename is your choice? Like, what you call the file? If so, maybe rephrase to "the name of the file is your choice." Just so it's clear :)

xml/networking-pcipt_sriov_config.xml
```diff
@@ -115,6 +115,115 @@
...
+       including Nvidia drivers.
+      </para>
+     </step>
+     <step>
+      <para>
+       On the host, <literal>iommu_groups</literal> should be enabled. To check
```
asettle: Probably a question for the OP, but is this insinuating that it is already enabled and you should check that? Or is it saying that you need to enable it now?

xml/networking-pcipt_sriov_config.xml
```diff
@@ -115,6 +115,115 @@
...
+       if IOMMU is enabled:
+      </para>
+<screen>&prompt.root;virt-host-validate
+.....
+QEMU: Checking if IOMMU is enabled by kernel
+: WARN (IOMMU appears to be disabled in kernel. Add intel_iommu=on to kernel cmdline arguments)
+.....</screen>
+      <para>
+       To modify the kernel cmdline as suggested in the warning, edit the file
```
asettle: Should we expand this to fully say either CLI or command line? Or do we normally say cmdline?

xml/networking-pcipt_sriov_config.xml
asettle: @sjamgade ^^

sjamgade: this is a kernel feature and user might have it enabled already, since its usage is not limited to OpenStack. 
But because that is needed for pci passthrough, this is line is mentioning that it needs to be enabled. Followed by how to check it.

xml/networking-pcipt_sriov_config.xml
csymons-suse: This usage is consistent with the immediately preceding : WARN message
: WARN (IOMMU appears to be disabled in kernel. Add intel_iommu=on to kernel cmdline arguments)

Should stay as it is

xml/networking-pcipt_sriov_config.xml
csymons-suse: changed

xml/networking-pcipt_sriov_config.xml
csymons-suse: clarified

PR #781, opened 2019-02-22
Remove Remnants of HPE Stuff

sknorr:
* replace all the old HOS entities
* remove 404'ing links to HPE docs
* remove those part of the Helion release notes that are _completely_ irrelevant


sknorr:
@csymons-suse @jeremy-moffitt Do either you know anything about HPE SSA (Smart Storage Administrator) and whether that is still available for SOC? (I commented one mention of it but there were more and I did not doing anything about those.)

KeithMnemonic:
HP SSA is a tool you access from the BIOS to provision disks/arrays. It is not tied to SOC but rather HPE HW. The HPSSA CLI is a linux tool if that is what you are referring to.  

https://support.hpe.com/hpsc/swd/public/detail?swItemId=MTX_e91ff24526b84cf2a502b27b41

csymons-suse:
What Keith wrote, plus it is generally available, and not necessarily tied to HOS, HPE, hLinux or RHEL. IIRC, the instructions cover that use case.

PR #792, opened 2019-02-27
Update log level instructions and defaults (bsc#1123231)

cadenzajon:


aojea:
LGTM but prefer if @mnmehta and @KeithMnemonic comment on it

PR #793, opened 2019-02-27
rabbitmq cluster replace node (SCRD-7468,SCRD-7469)

csymons-suse:


xml/operations-troubleshooting-recover_rabbit.xml
```diff
@@ -204,29 +204,68 @@ ardana-cp1-c1-m3-mgmt</screen>
...
+    <para>
+     Force the cluster to forget the node you are removing (in this example,
+     the controller 3 node).
+    </para>
+    <screen>&prompt.ardana;sudo rabbitmqctl forget_cluster_node \
+rabbit@cloud-cp1-rmq-mysql-m3-mgmt</screen>
+   </step>
+   <step>
+    <para>
+     Confirm that the node has been removed:
```
dmpop: _has been removed._ (period instead of colon)

xml/operations-troubleshooting-recover_rabbit.xml
```diff
@@ -204,29 +204,68 @@ ardana-cp1-c1-m3-mgmt</screen>
    baremetal node, the &rabbit; cluster needs any references to the node
    removed from it. This is because &rabbit; associates a cookie with each node
    in the cluster which is derived, in part, by the specific hardware.
-   Replacing a hard drive in a node should be fine, but an exchanged
-   motherboard or replacing it with another node entirely will potentially
-   trigger a breakage. Under these circumstances, the running &rabbit; cluster
-   must be edited from a running &rabbit; node with these steps:
+   Replacing a hard drive in a node should be fine, but changing a motherboard
```
dmpop: In general, we avoid using _should_ and _would_ in technical documentation as this introduces ambiguity. It's either fine or it's not. Also, _trigger a breakage_ sounds clunky. Maybe something like this?

_While it's possible to replace a hard drive without problems, swapping a motherboard may cause interruption._

csymons-suse: Thank you for proofreading and copy editing. Fixed and then some. The whole paragraph was clunky.

xml/operations-troubleshooting-recover_rabbit.xml
csymons-suse: done

PR #794, opened 2019-02-27
rabbitmq cluster replace node (SCRD-7468,SCRD-7469)

csymons-suse:


csymons-suse:
same content as https://github.com/SUSE-Cloud/doc-cloud/pull/793
screenshot at 793

PR #797, opened 2019-03-01
Add references to DIB back to guide

cmurphy:
This partially reverts #761. It's not forbidden to mention
diskimage-builder or other freely licensed distros, so long as we call
out that special customizations to ramdisk images or boot ISOs can't be
supported. This also corrects an error that #761 had removed, which
called out 'suse' as a DIB element. 'suse' is not an element available
in DIB (it's 'opensuse'), and moreover the section called for a special
'-signed' element which is not available for opensuse, so we need to
reference the 'ubuntu-signed' element.

### DIFF
--- a/xml/admin-ironic.xml
+++ b/xml/admin-ironic.xml
@@ -1941,7 +1941,14 @@ nova boot --flavor ironic-test-3 --image test-image instance-1</screen>
                             machine that matches with user specified flavors.</para>
                     </listitem>
                   </itemizedlist>
-                  <para>The automatic boot ISO creation for UEFI boot mode has been enabled in Kilo.</para>
+                  <para>The automatic boot ISO creation for UEFI boot mode has been enabled in Kilo.
+                    The manual creation of boot ISO for UEFI boot mode is also possible, but note that it is not officially supported.
+                    For the latter, the boot ISO for the deploy image needs to be built
+                    separately and the deploy image’s <literal>boot_iso</literal> property in glance should
+                    contain the glance UUID of the boot ISO. For building boot ISO, add the <literal>iso</literal>
+                    element to the <link xlink:href="https://pypi.python.org/pypi/diskimage-builder">diskimage-builder</link>
+                    command to build the image. For example:</para>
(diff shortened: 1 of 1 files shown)


xml/admin-ironic.xml
```diff
@@ -1941,7 +1941,13 @@ nova boot --flavor ironic-test-3 --image test-image instance-1</screen>
                             machine that matches with user specified flavors.</para>
                     </listitem>
                   </itemizedlist>
-                  <para>The automatic boot ISO creation for UEFI boot mode has been enabled in Kilo.</para>
+                  <para>The automatic boot ISO creation for UEFI boot mode has been enabled in Kilo.
+                    The manual creation of boot ISO for UEFI boot mode is also possible, but note that it is not officially supported.
+                    For the latter, the boot ISO for the deploy image needs to be built
+                    separately and the deploy image’s <literal>boot_iso</literal> property in glance should
+                    contain the glance UUID of the boot ISO. For building boot ISO, add the <literal>iso</literal>
+                    element to the diskimage-builder command to build the image. For example:</para>
```
skazi0: maybe add a link to DIB as below?

xml/admin-ironic.xml
```diff
@@ -1973,6 +1979,14 @@ nova boot --flavor ironic-test-3 --image test-image instance-1</screen>
                     specified flavors but deployment would not use its secure boot capability.
                     Secure boot deploy would happen only when it is explicitly specified through
                     flavor.</para>
+                  <para>You may use element <literal>ubuntu-signed</literal> or <literal>fedora</literal> to build signed deploy iso and
+                    user images with <link xlink:href="https://pypi.python.org/pypi/diskimage-builder">diskimage-builder</link>, but note that custom images cannot be officially supported.
```
skazi0: s/cannot/are not/ ?

xml/admin-ironic.xml
cmurphy: done

xml/admin-ironic.xml
cmurphy: done

PR #799, opened 2019-03-04
Fixing broken ref

asettle:
The package build was broken due to the wrong ref being implemented. This PR changes that.

See error logs: https://ci.suse.de/view/Cloud/view/Meta/job/cloud-trackupstream/component=documentation-suse-openstack-cloud,label=openstack-trackupstream,project=Devel%3ACloud%3A8%3AStaging/1826/console


asettle:
Sorry @csymons-suse must have missed this when reviewing your PR earlier. My understanding is that this is the correct variant as per here: https://github.com/SUSE-Cloud/doc-cloud/blob/541f658202281f45bad897a8056e040e28baee81/xml/entity-decl.ent#L32

Let me know if it should be something different!

PR #800, opened 2019-03-05
Fix Novdoc conversion issues in Upstream Admin guide

sknorr:
via http://docserv.nue.suse.com/online-docs-reports/report-2019-03-02_18-01

@asettle / @dpopov could you specifically check the last commit that removes the empty sections for whether it does (additional) harm or is benign? I am not sure if these sections have always been empty and didn't want to investigate.

### DIFF
--- a/xml/admin-ironic.xml
+++ b/xml/admin-ironic.xml
@@ -3575,9 +3575,6 @@ URL</literal>.</para>
                   <screen>ironic port-create -n $NODE -a &lt;MAC_address_of_Ucs_server's_NIC&gt;</screen>
                 </section>
               </section>
-              <section>
-                <title>References</title>
-              </section>
             </section>
           </section>
           <section xml:id="cimc-driver">
--- a/xml/admin-ironic.xml
+++ b/xml/admin-ironic.xml
@@ -1948,7 +1948,7 @@ nova boot --flavor ironic-test-3 --image test-image instance-1</screen>
                     contain the glance UUID of the boot ISO. For building boot ISO, add the <literal>iso</literal>
                     element to the <link xlink:href="https://pypi.python.org/pypi/diskimage-builder">diskimage-builder</link>
                     command to build the image. For example:</para>
-￼                   <screen>disk-image-create ubuntu baremetal iso</screen>
+                   <screen>disk-image-create ubuntu baremetal iso</screen>
                 </section>
                 <section xml:id="ilo-uefi-secure-boot-support">
                   <title>UEFI Secure Boot Support</title>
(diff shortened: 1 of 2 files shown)


xml/admin-neutron-config-az.xml
```diff
@@ -292,9 +292,6 @@ $ openstack network agent show 9632309a-2aa4-4304-8603-c4de02c4a55f
                         capacity left in the zone requested.</para>
       </note>
     </section>
-    <section>
-      <title>Availability zone aware scheduler</title>
-    </section>
```
asettle: So this shouldn't be removed, it's just not formatting quite right. This needs to be the section title, and the Network Scheduler needs to be the sub title within that. See: https://docs.openstack.org/newton/networking-guide/config-az.html

xml/admin-ironic.xml
```diff
@@ -3575,9 +3575,6 @@ URL</literal>.</para>
                   <screen>ironic port-create -n $NODE -a &lt;MAC_address_of_Ucs_server's_NIC&gt;</screen>
                 </section>
               </section>
-              <section>
-                <title>References</title>
-              </section>
```
asettle: This is missing a link: https://docs.openstack.org/ironic/pike/admin/drivers/ucs.html#references

xml/admin-keystone.xml
```diff
@@ -83729,9 +83729,6 @@ class Something(object):
             </varlistentry>
           </variablelist>
         </section>
-        <section xml:id="module-keystone_module-contents">
-          <title>Module contents</title>
```
asettle: This one is particular weird. These sections should belong to the API content. I'll investigate further, this one is weird.

xml/admin-neutron-config-az.xml
asettle: Ugh hang on, that link is too many versions old.

asettle: https://docs.openstack.org/neutron/pike/admin/config-az.html#availability-zone-aware-scheduler

xml/admin-keystone.xml
dmpop: So is it weird or is it weird weird?

asettle: Weird weird. It's all API config content. How did that get converted with the admin guide? Removed here:  https://github.com/SUSE-Cloud/doc-cloud/pull/800/commits/4f79d3f128c377fe3429c2ea12c961b656ae1124

asettle: Basically, everything from here: https://docs.openstack.org/keystone/pike/configuration/index.html
Got converted too.

xml/admin-ironic.xml
asettle: Although relaly this one doesn't matter

xml/admin-neutron-config-az.xml
asettle: This one should be better nested

asettle:
@sknorr I think that last PR deals with it :) the other one can be deleted.

sknorr:
Thanks a lot @asettle !

PR #708, opened 2019-01-11
Removing the "User Guide" (cloud admin)

asettle:
This PR removes the (titled) User Guide for CLM which contains content pertaining to Cloud Administration and Project User Guide. This differs from the End User Guide for Crowbar.
This guide will be replaced by the Upstream User Guides.
The Ops Console content that is in the Cloud Admin section of this guide has been moved out to the Operations Guide. See this PR https://github.com/SUSE-Cloud/doc-cloud/pull/707

PR #808, opened 2019-03-06
Migrating cloud admin and user guide to ops guide

asettle:
This PR adds administration and user content from the previously titled User Guide for CLM into the Operations Guide for CLM.

This content is being migrated as the new SOC9 structure favors the usage of the upstream Administration and User Guides. Migrating the content to the CLM Operations Guide ensures that we do not lose CLM-specific user and admin content. 

The User Guide was previously deleted in this PR: https://github.com/SUSE-Cloud/doc-cloud/pull/708

### DIFF
--- a/xml/book.operations.xml
+++ b/xml/book.operations.xml
@@ -10,17 +10,12 @@
  <info>
   <abstract>
    <para>
-    This guide provides a list of useful procedures for managing your
-    &product; cloud. The audience is the admin-level operator of the
-    cloud.
+     At the time of the &productname; &productnumber; release, this guide
+     contains information pertaining to the operations,
+     administration, and user functions of &productname;.
+     The audience is the admin-level operator of the cloud.
    </para>
-   <important>
-     <para>
-       At the time of the &productname; &productnumber; release, this guide
-       contains information pertaining to the operations,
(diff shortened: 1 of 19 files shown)


xml/book.operations.xml
```diff
@@ -10,10 +10,12 @@
  <info>
   <abstract>
    <para>
-    This guide provides a list of useful procedures for managing your
-    &product; cloud. The audience is the admin-level operator of the
-    cloud.
+     At the time of the &productname; &productnumber; release, this guide
+     contains information pertaining to the operations,
+     administration, and user functions of &productname;.
+     The audience is the admin-level operator of the cloud.
```
asettle: One thing I want to draw your attentions to @dmpop and @csymons-suse is this. I changed the intro so that our users are aware that this guide doesn't *just* contain operations content (which is quite an important factor with the upstream stuff we're using). Let me know how you feel on wording etc

csymons-suse: I like it. Maybe "operations" > "operation". Sounds better when I read it aloud.

asettle: Yeah that makes sense! Thanks Carl :) Will update.

PR #812, opened 2019-03-07
change _member_, Member roles (SCRD-7612)

csymons-suse:


xml/book_cloud_admin_crowbar.xml
```diff
@@ -36059,9 +36059,9 @@ User has been updated.</screen>
 +----------------------------------+---------------+
 | ID                               | Name          |
 +----------------------------------+---------------+
-| 71ccc37d41c8491c975ae72676db687f | Member        |
+| 71ccc37d41c8491c975ae72676db687f | member        |
 | 149f50a1fe684bfa88dae76a48d26ef7 | ResellerAdmin |
-| 9fe2ff9ee4384b1894a90878d3e92bab | _member_      |
+| 9fe2ff9ee4384b1894a90878d3e92bab | member        |
```
mbelur: we don't need two rows showing the same role...

csymons-suse: fixed. Thank you for catching that. Slow down and think, buddy.

PR #836, opened 2019-03-21
remove enabling jumbo frames (SCRD-8025)

csymons-suse:


csymons-suse:
overdid it. Be back with less.

PR #837, opened 2019-03-22
remove ceilometer-api, ceilometer-client (SCRD-7672, SCRD-8476)

csymons-suse:


xml/metering-metering_metertypes.xml
```diff
@@ -116,15 +116,9 @@
...
-  <para>
-   Detailed information on the Ceilometer API can be found on the following
-   page:
-  </para>
-  <para>
-   <link xlink:href="http://docs.openstack.org/developer/ceilometer/webapi/v2.html">Ceilometer
-   Web API</link>.
+   installation. More information about Ceilometer can be found at <link
+   xlink:href="http://docs.openstack.org/developer/ceilometer/webapi/v2.html">OpenStack
+   Ceilometer</link>.
```
jodavis-suse: This redirects to https://docs.openstack.org/ceilometer/latest/  Would that be a better url to use?

xml/metering-metering_overview.xml
```diff
@@ -9,24 +9,16 @@
...
-   the meters provided by all the OpenStack services. The data collected
-   is available both through the Monasca API and the Ceilometer V2 API.
- </para>
- <important>
-   <para>
-     Ceilometer V2 API was deprecated in the Pike release upstream.  Although
-     the Ceilometer V2 API is still available with &productname;, to prepare for
-     eventual removal of Ceilometer V2 API, we recommend that users switch to
-     the Monasca API to access data.
+   Ceilometer acts as the central collection and data access service to the
```
jodavis-suse: Lost the indentation here.  I assume that doesn't matter for the xml, just looked odd that it now lines up with <para> instead of being two spaces right of it.

csymons-suse: strange...looks okay in the source file.

xml/metering-metering_metertypes.xml
csymons-suse: Thanks for checking that...better. fixed and force-pushed.

xml/metering-metering_reconfig.xml
```diff
@@ -346,16 +171,15 @@ Jun 12 05:07:14 ardana-qe201-cp1-c1-m2-mgmt systemd[1]: Started ceilometer-agent
...
-   state. &productname; does not monitor these processes anymore so the best option to
-   resolve this issue is to manually delete alarms that are no longer used but
-   are installed.
+   The Monasca HTTP Process monitors Ceilometer's notification and polling
+   agents are monitored. If these agents are down, Monasca monitoring alarms
+   are triggered. You can use the notification alarms to debug the issue and
+   restart the notifications agent. However, for Central-Agent (polling) and
+   Collector the alarms need to be deleted. These two processes are not started
+   after an upgrade so when the monitoring process checks the alarms for these
+   components, they will be in UNDETERMINED state. &productname; does not
```
asettle: should UNDERERMINED be marked up with <literal> ?

PR #845, opened 2019-03-26
add explanation, add global_physnet_mtu parameter (SCRD-8025)

csymons-suse:
in neutron.conf.j2, advertise_mtu has been deprecated


xml/networking-configure_mtu.xml
```diff
@@ -167,18 +169,33 @@
...
+...
+global_physnet_mtu = 9000</screen>
+    <para>
+     This allows Neutron to advertise the optimal MTU to instances (based upon
+     global_physnet_mtu minus the encapsulation size).
+    </para>
+   </step>
+   <step>
+    <para>
+     Remove the "dhcp-option-force=26,1400" line from
```
asettle: Should this be marked up into literal also? I'd suggset so: <literal>dhcp-option-force=26,1400</literal>

PR #855, opened 2019-03-28
Isolated Metadata Settings (SCRD-8557)

aojea:
In order to disable metadata in the dhcp namespaces you have to disable
enable_isolated_metadata = False
force_metadata = False

Previously, force_metadata was always false, so it never was documented that we have to change it.
Currently, there is a situation when we are enabling it, hence we need to document it.

### DIFF
--- a/xml/networking-isolated_metadata.xml
+++ b/xml/networking-isolated_metadata.xml
@@ -11,20 +11,23 @@
   In &productname;, Neutron currently sets <literal>enable_isolated_metadata =
   True</literal> by default in <literal>dhcp_agent.ini</literal> because
   several services require isolated networks (Neutron networks without a
-  router). This has the effect of spawning a neutron-ns-metadata-proxy process
-  on one of the controller nodes for every active Neutron network.
+  router). It also sets <literal>force_metadata = True</literal> if DVR is 
+  enabled to improve the scalability on large environments with a high churn 
+  rate. However, this has the effect of spawning a neutron-ns-metadata-proxy
+  process on one of the controller nodes for every active Neutron network.
  </para>
  <para>
   In environments that create many Neutron networks, these extra
   <literal>neutron-ns-metadata-proxy</literal> processes can quickly eat up a
-  lot of memory on the controllers, which does not scale well.
+  lot of memory on the controllers, which does not scale up well.
  </para>
  <para>
   For deployments that do not require isolated metadata (that is, they do not
   require the Platform Services and will always create networks with an
-  attached router), you can set <literal>enable_isolated_metadata =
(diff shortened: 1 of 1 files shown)


xml/networking-isolated_metadata.xml
```diff
@@ -11,20 +11,23 @@
   In &productname;, Neutron currently sets <literal>enable_isolated_metadata =
   True</literal> by default in <literal>dhcp_agent.ini</literal> because
   several services require isolated networks (Neutron networks without a
-  router). This has the effect of spawning a neutron-ns-metadata-proxy process
-  on one of the controller nodes for every active Neutron network.
+  router). It also sets <literal>force_metadata = True</literal> if DVR is 
+  enabled to improve the scalability on large environments with a high churn 
+  rate. However, this has the effect of spawning a neutron-ns-metadata-proxy
```
asettle: <literal>neutron-ns-metadata-proxy</literal>

xml/networking-isolated_metadata.xml
```diff
@@ -11,20 +11,23 @@
...
  </para>
  <para>
   For deployments that do not require isolated metadata (that is, they do not
   require the Platform Services and will always create networks with an
-  attached router), you can set <literal>enable_isolated_metadata =
-  False</literal> in dhcp_agent.ini to reduce Neutron memory usage on
-  controllers, allowing a greater number of active Neutron networks.
+  attached router) and do not have a high churn rate, you can set 
+  <literal>enable_isolated_metadata = False</literal> and <literal>force_metadata = False</literal>
+  in dhcp_agent.ini to reduce Neutron memory usage on controllers,
```
asettle: <literal>dhcp_agent.ini</literal>

xml/networking-isolated_metadata.xml
```diff
@@ -11,20 +11,23 @@
   In &productname;, Neutron currently sets <literal>enable_isolated_metadata =
```
rktidwell: "Neutron currently sets" - this isn't exactly true, neutron defaults this to false. SUSE cloud sets it to true. https://docs.openstack.org/neutron/rocky/configuration/dhcp-agent.html

aojea: @rktidwell , you missed the first part of the sentence that explains exactly that 

 ```In &productname;, Neutron currently sets ```

aojea:
@rktidwell I want to keep this  PR only to address the bug mentioned, please create new bugs for other issues or enhancements
@asettle please review it again

xml/networking-isolated_metadata.xml
rktidwell: I did misunderstood that. Thanks for setting me straight, this all makes perfect sense now.

PR #856, opened 2019-03-28
Replace postgres with mariadb (bz#1130795)

aojea:
Crowbar default database is MariaDB since Cloud8


aojea:
don't know your workflow @asettle , should I merge or wait until one of you merge it?

asettle:
@aojea we don't really have anything set in place formally, but if one of us has approved, feel free to merge your own patches in future :)
In this case, I'll go ahead and merge now! Thanks for this :)

PR #865, opened 2019-03-29
Update migration guide to incorporate Carl's changes (SCRD-3763)

cadenzajon:


xml/art_migration.xml
```diff
@@ -1617,8 +1617,7 @@ fi</screen>
      &ostack-current;. When running &ostack; commands an operator may see
      messages saying <literal>Policy doesn't allow ...</literal> such as:
     </para>
-    <screen>Policy doesn't allow volume_extension:services:index to be
-    performed</screen>
+    <screen>Policy doesn't allow volume_extension:services:index to be performed</screen>
```
asettle: Should volume_extension:services:index be marked up even in the screen? I imagine it would still be marked up in an error message.

csymons-suse: I don't think it should be tagged; it's an example of an error message in terminal.

PR #883, opened 2019-04-03
update swift CLI to OSC (SCRD-7647)

csymons-suse:


xml/operations-maintenance-controller-full_recovery_test.xml
```diff
@@ -46,11 +46,11 @@
...
    <para>It used 5 servers: 3 Controllers and 2 computes.</para>
-   <para>The controller node have three disks. The first one is reserved for the system, while others are used for swift.</para>
+   <para>The controller node have three disks. The first one is reserved for the system, while others are used for &o_objstore;.</para>
    <note>
-     <para>During this Disaster Recovery exercise, we have saved the data on disk 2 and 3 of the swift controllers.</para>
-     <para>This allow to restore the swift objects after the recovery.</para>
-     <para>If these disks were to be wiped as well, swift data would be lost but the procedure would not change.</para>
+     <para>During this Disaster Recovery exercise, we have saved the data on disk 2 and 3 of the &o_objstore; controllers.</para>
+     <para>This allow to restore the &o_objstore; objects after the recovery.</para>
+     <para>If these disks were to be wiped as well, &o_objstore; data would be lost but the procedure would not change.</para>
```
asettle: Can all this text above be wrapped so it fits in the window? :) not a blocker, obvs. But just helpful for later!

xml/operations-objectstorage-swift_cli.xml
```diff
@@ -20,8 +22,9 @@
   Then you can use your preferred editor to edit swiftuser.osrc so you can
   authenticate using the <literal>OS_USERNAME</literal>,
   <literal>OS_PASSWORD</literal>, and <literal>OS_PROJECT_NAME</literal> you
-  wish to use. For example, if you would like to use the <literal>demo</literal>
-  user that is created automatically for you, then it might look like this:
+  wish to use. For example, if you would like to use the
```
asettle: you want to use*

xml/operations-objectstorage-swift_cli.xml
```diff
@@ -20,8 +22,9 @@
...
+  <literal>demo</literal> user that is created automatically for you, then it
+  might look like this:
```
asettle: remove "then it might look like this" - it works with just "for example" :) Or maybe switch so it reads...
"The following example displays the configuration if you want to use the demo user that is created automatically for you:"

PR #884, opened 2019-04-03
clean up DPDK sections (SCRD-7774)

csymons-suse:


xml/networking-ts_dpdk.xml
```diff
@@ -177,9 +162,7 @@ NUMA node1 CPU(s):     12-23,36-47</screen>
    </listitem>
    <listitem>
     <para>
-     Jumbo Frames MTU does not work with DPDK OvS. There is an upstream patch
-     most likely showing up in OvS 2.6 and it cannot be backported due to
-     changes this patch relies upon.
+     Jumbo Frames MTU is not supported with DPDK.
```
spacefito: We have to state this until we can verify otherwise. This statement is not being changed anyways, is just being cleaned up. so

PR #888, opened 2019-04-03
Fix command for listing compute hosts (SCRD-7602)

jeremy-moffitt:
Section 6.1.1 had an outdated command in step 5 for listing hosts running compute services, updated to currently used command

### DIFF
--- a/xml/operations-compute-creating_aggregates.xml
+++ b/xml/operations-compute-creating_aggregates.xml
@@ -91,11 +91,10 @@
    <listitem>
     <para>
      Next, you need to add compute hosts to this aggregate so you can start by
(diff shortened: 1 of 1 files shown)


csymons-suse:
openstack replacement table
https://docs.openstack.org/python-openstackclient/latest/cli/decoder.html#nova-cli

jeremy-moffitt:
@csymons-suse in this case the original output was being piped to a grep for compute to narrow the host list down to only hosts running the compute services... `hypervisor list` performs that filtering internally, so the entirety of `openstack host list | grep compute` is replaced by `openstack hypervisor list` without the need to manually grep the results

PR #894, opened 2019-04-05
Remove non-functional DC file

sknorr:
Not urgent but this confused me.


rsalevsky:
We should also remove the upstream-admin and upstream-user guide as they come from a different source going forward.

asettle:
Yes this seems about right. Adding @csymons-suse and @jeremy-moffitt to the review too.

PR #895, opened 2019-04-05
Rename guides to match SOC naming scheme

rsalevsky:


asettle:
Checked build for all 3 guides. A-okay.

PR #896, opened 2019-04-06
add --volume-type <type> to OSC volume create (SCRD-8021)

csymons-suse:


xml/bk_openstack_user.xml
```diff
@@ -3824,7 +3824,7 @@ which you want to create your volume:</para>
           <step>
             <para>Create a volume with 8 gibibytes (GiB) of space, and specify the
 availability zone and image:</para>
-            <screen>$ openstack volume create --image 8bf4dc2a-bf78-4dd1-aefa-f3347cf638c8 \
+            <screen>$ openstack volume create --volume-type <replaceable>TYPE</replaceable> --image 8bf4dc2a-bf78-4dd1-aefa-f3347cf638c8 \
```
guangyee: --volume-type should be --type

csymons-suse: all "--volume-type" occurrences changed to "--type"
added note about "openstack volume type list" command

PR #898, opened 2019-04-08
Add warning about Day2 disconnects with self-signed certs (SCRD-6373)

jeremy-moffitt:
Add a warning block in the configuration section of the Day 2 docs that informs users that if they're using self-signed certificates, their streaming configuration updates may disconnect. References the security guide (where there are instructions for using signed certs)

### DIFF
--- a/xml/day2_ui_documentation.xml
+++ b/xml/day2_ui_documentation.xml
@@ -163,7 +163,7 @@
  <title>Configuration</title>
  <para>
   The <guimenu>Configuration</guimenu> tab displays services that
-  are deployed in the cloud and the configuration files associated with thosei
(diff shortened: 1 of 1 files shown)


eapodaca:
would it be possible to link to the security guide instead of simply mentioning it?

jeremy-moffitt:
> would it be possible to link to the security guide instead of simply mentioning it?

Ideally yes, I'm not sure if we have that capability though, will need one of the doc writers to chime in... particularly complicated by the fact that we deliver these docs multiple ways... but I do think that would be a better solution

jeremy-moffitt:
changed from warning to note, and linked the specific section of the security guide

PR #899, opened 2019-04-08
Update Day2 access reqs to include keystone and HAProxy (SCRD-8513)

jeremy-moffitt:
HAProxy and Keystone are required to access the Day2 UI, updated the access paragraph to indicate as much

### DIFF
--- a/xml/day2_ui_documentation.xml
+++ b/xml/day2_ui_documentation.xml
@@ -36,7 +36,9 @@
   was deployed. Access to this network is necessary to be able to access the
   &clm; Admin UI and log in. Depending on the network setup, it may be
   necessary to use an SSH tunnel similar to what is recommended in <xref
-  linkend="running_install_ui"/>.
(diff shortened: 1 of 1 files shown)


xml/day2_ui_documentation.xml
```diff
@@ -36,7 +36,9 @@
   was deployed. Access to this network is necessary to be able to access the
   &clm; Admin UI and log in. Depending on the network setup, it may be
   necessary to use an SSH tunnel similar to what is recommended in <xref
-  linkend="running_install_ui"/>.
+  linkend="running_install_ui"/>. The Admin UI requires Keystone and HAProxy
+  to be running and to be accesible. If Keystone or HAProxy are not running,
```
csymons-suse: Keystone -> `&o_ident;`; HAProxy -> &haproxy;

seems like Cloud should not be capitalized

jeremy-moffitt: good catch, fixed in latest revision. thanks!

PR #902, opened 2019-04-08
Fix mysql TLS enablement comment (SCRD-8757)

cadenzajon:


xml/security-tls_mysql.xml
```diff
@@ -102,7 +102,7 @@ mysql_gcomms_bind_tls: False</screen>
     <para>
      Uncomment the appropriate line so the file looks like this:
     </para>
-<screen># TLS disabled for cluster
+<screen># TLS enabled for cluster
```
csymons-suse: shouldn't `# TLS enabled for cluster` be uncommented in order to enable TLS?

cadenzajon: Definitely not. It is correct as-is.

xml/security-tls_mysql.xml
```diff
@@ -102,7 +102,7 @@ mysql_gcomms_bind_tls: False</screen>
...
 mysql_gcomms_bind_tls: "{{ host.bind['FND_MDB'].mysql_gcomms.tls }}"
```
asettle: Ah, is this the line that is supposed to be uncommented? Makes more sense now.

PR #906, opened 2019-04-08
Add final step to update /etc/hosts when removing compute node

cadenzajon:


xml/operations-maintenance-compute-remove_compute_node.xml
```diff
@@ -375,22 +376,19 @@ $ neutron agent-delete f0d262d1-7139-40c7-bdc2-f227c6dee5c8</screen>
...
-ansible-playbook -i hosts/localhost config-processor-run.yml</screen>
-    <para>
-     To free up the resources when running the configuration processor, use
+     To release the network capacity allocated to the deleted server(s), use
      the switches <literal>remove_deleted_servers</literal> and
-     <literal>free_unused_addresses</literal>. For more information, see
-     <xref linkend="persisteddata"/>.
+     <literal>free_unused_addresses</literal> when running the configuration
+     processors. (For more information, see
+     <xref linkend="persisteddata"/>.)
```
asettle: Probably doesn't need to be in parenthesis?

xml/operations-maintenance-compute-remove_compute_node.xml
```diff
@@ -399,6 +397,14 @@ ansible-playbook -i hosts/localhost config-processor-run.yml -e remove_deleted_s
 <screen>cd ~/openstack/ardana/ansible
 ansible-playbook -i hosts/localhost ready-deployment.yml</screen>
    </step>
+   <step>
+    <para>
+     Refresh the /etc/hosts file through the cloud to remove references to the
```
asettle: <literal>/etc/hosts</literal>

xml/operations-maintenance-compute-remove_compute_node.xml
cadenzajon: Certainly it should, to reduce its weight in the paragraph and tighten the correlation between the previous statement (about running the configuration processor) and the following statement (the <screen> list of commands to do so).

PR #917, opened 2019-04-12
Remove documentation about Cisco UCS Crowbar integration

vuntz:
This has been hidden in Crowbar, because it's not used and not tested.


asettle:
@vuntz merge at will

PR #957, opened 2019-04-24
Fix bsc#1118003

dmpop:


xml/operations-change_service_passwords.xml
```diff
@@ -87,8 +87,9 @@
   </para>
   <important>
    <para>
-    Currently, you can not use any special characters with Ansible Vault,
-    Service Passwords, or vCenter configuration.
+    It is possible to use special characters in passwords. However, the
+    <literal>$</literal> character must be escaped by entering it twice. For
```
asettle: Must be escaped? Haven't heard that before. Makes sense with the example, but wondering if there's a better way to explain this. I admittedly cannot think of anything, so I'm sure this is fine.

PR #977, opened 2019-05-14
Fix documentation to setup additional external networks

vuntz:
This is based on customer feedback.

@rhafer I put you as reviewer for the second commit, as I could be misunderstanding something here...


vuntz:
Btw, it would be good to backport to Cloud 8 too.

PR #981, opened 2019-05-15
clarify LDAP manual vs GUI (bsc#1134495)

csymons-suse:
missing reference to deploy documentation. xref linkend is not working as expected.

### DIFF
--- a/xml/admin-keystone.xml
+++ b/xml/admin-keystone.xml
@@ -1406,13 +1406,28 @@ domain_name_url_safe = off</screen>
                     value as the root user:</para>
             <screen language="console"># setsebool -P authlogin_nsswitch_use_ldap on</screen>
           </important>
-          <para>The Identity configuration is split into two separate back ends; identity
-                (back end for users and groups), and assignments (back end for domains,
-                projects, roles, role assignments). To configure Identity, set options
-                in the <literal>/etc/keystone/keystone.conf</literal> file. See
-                <link xmlns:xl="http://www.w3.org/1999/xlink" xl:href="#integrate-identity-backend-ldap">Integrate Identity back end with LDAP</link> for Identity back end configuration
-                examples. Modify these examples as needed.</para>
-          <para>
+          <para>The Identity configuration is split into two separate back
+          ends; identity (back end for users and groups), and assignments (back
+          end for domains, projects, roles, role assignments). To configure
+          Identity, set options in the
+          <filename>/etc/keystone/keystone.conf</filename> file. See <link
+          xmlns:xl="http://www.w3.org/1999/xlink"
+          xl:href="#integrate-identity-backend-ldap">Integrate Identity back
+          end with LDAP</link> for Identity back end configuration
+          examples. Modify these examples as needed.</para>
+                <para>
+                 <important>
+                  <para>
+                   Manual changes to
+                   <filename>/etc/keystone/keystone.conf</filename> or
+                   <filename>/etc/keystone/domains/keystone.DOMAIN_NAME.conf</filename>
+                   override entries made with the Keystone Barclamp (see the
+                   Crowbar Deployment Guide section on Deploying
+                   Keystone). Incorrect manual entries may cause errors. Some
+                   LDAP integration entries are defaulted in the Keystone
+                   Barclamp, so manual entries may not be necessary.
+                  </para>
+                 </important>
             <emphasis role="bold">To define the destination LDAP server</emphasis>
           </para>
           <para>Define the destination LDAP server in the <literal>/etc/keystone/keystone.conf</literal> file:</para>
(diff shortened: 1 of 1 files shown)


xml/admin-keystone.xml
```diff
@@ -1406,14 +1406,31 @@ domain_name_url_safe = off</screen>
                     value as the root user:</para>
```
cdevitasuse: Is there a section where it takes you thru the barclamp configuration? The edits that are made to the barclamp configuration require edits to the json file in raw mode. 
>
   "domain_specific_config": {
    "ldap_users": {
      "identity": {
        "driver": "ldap"
      },
      "ldap": {
        "url": "ldaps://ldap.suse.de",
        "user": "dc=Manager,dc=example,dc=com",
        "password": "",
        "suffix": "dc=suse,dc=de",
        "page_size": 0,
        "alias_dereferencing": "default",
        "query_scope": "one",
        "user_tree_dn": "ou=accounts,dc=suse,dc=de",
        "user_filter": "",
        "user_objectclass": "posixAccount", 
`

xml/admin-keystone.xml
```diff
@@ -1821,11 +1838,123 @@ user_enabled_default = 512</screen>
...
+             unique options are permissible in each of the Keystone
+             configuration files:
+            </para>
+            <screen>/etc/keystone/keystone.conf
+/etc/keystone/domains/keystone.<replaceable>DOMAIN</replaceable>.conf
+/etc/keystone/keystone.d/010-keystone.conf
+/etc/keystone/keystone.d/100-keystone.conf
+/etc/keystone/keystone.d/999-keystone.conf</screen>
+            <para>
+             If there are duplicate entries, there will be no output in the log
```
asettle:
```suggestion
If there are duplicate entries, there is no output in the log
```

xml/admin-keystone.xml
```diff
@@ -1821,11 +1838,123 @@ user_enabled_default = 512</screen>
...
+             file <filename>/var/log/keystone/keystone.log</filename>. In
+             addition, the following error message will appear in
```
asettle:
```suggestion
addition, the following error message appears in
```

xml/admin-keystone.xml
```diff
@@ -1821,11 +1838,123 @@ user_enabled_default = 512</screen>
...
+             </step>
+            </procedure>
+           </listitem>
+          </itemizedlist>
+         </section>
       </section>
       <section xml:id="upgrading-keystone" xml:base="identity-upgrading">
         <title>Upgrading Keystone</title>
-        <para>As of the Newton release, keystone supports two different approaches to
+        <para>As of the Newton release, Keystone supports two different approaches to
```
asettle: So this is something I'd like some clarity on. I don't think we have it defined in the style guide, but upstream we have uppercase for the service name (aka: Identity service) and lowercase for the project names (keystone), see: https://docs.openstack.org/doc-contrib-guide/writing-style/openstack-components.html

Do we have this defined or has this been discussed before @csymons-suse @dmpop? Might be worthwhile to sync with upstream.

asettle: Just checking in with TR :)

asettle: Okay just quickly checked in with TR, no major opinions but preference to follow the community guidelines (which, FTR, I recognise is completely different to capitalising the proper noun)

csymons-suse: I just work here =B^) However, if we're really part of the OpenStack Community then we should be consistent with it. 

Actually I never thought about this. I've only used the entity references...all capitalized. So it would seem that it has been discussed before.

Might be a bit of a project to clean this up...
- changing the entity references is simple
- but also correcting the occurrences that are not entities
- probably something that should be done anyway
- my inclination is to keep pushing towards consistency with upstream, based on comments from SUSE decision makers

asettle: You're right, that would be a lot. Perhaps we create it as a stretch goal for our Cloud 10 clean up? Thoughts @csymons-suse and @dmpop? We can at least ensure that SOC10 is correct, and from there work backwards whenever we notice something and have a JIRA card open for long-term tracking.

asettle: Thinking outloud, the entities shouldn't be too hard (as you said) but we should be able to perform a find and replace. Although I say that as if we have infinite time to fix these things between the three of us. Either way, let me know what you think and I'm happy to go ahead and create a JIRA card with the three of us as watchers and add this to the Cloud 10 spec.

xml/admin-keystone.xml
csymons-suse: changed

xml/admin-keystone.xml
csymons-suse: changed

xml/admin-keystone.xml
csymons-suse: keystone not changed to initial capital. Many occurrences of lc initial letter.

PR #1007, opened 2019-07-10
update MariaDB manually with CLI (bsc#1132852, SOC-9022)

csymons-suse:


asettle:
Same comments on https://github.com/SUSE-Cloud/doc-cloud/pull/1055

xml/depl_maintenance.xml
```diff
@@ -1816,14 +1816,13 @@ done</screen>
    10.2.22 Release Notes - Notable Changes</link>.
   </para>
   <para>
-   Update MariaDB with the following procedure:
+   Using the Pacemaker GUI, update MariaDB with the following procedure:
```
dmpop: There is an entity for Pacemaker.

csymons-suse: No entity for Pacemaker in Cloud 7 version of entity-decl.ent

asettle: That is true. Even so, it's literally "pacemaker": https://github.com/SUSE-Cloud/doc-cloud/blob/maintenance/cloud_9/xml/entity-decl.ent#L115

We could add it into cloud 7 but I imagine we're not going to be updating Cloud 7 that often anyway

PR #1020, opened 2019-07-11
Fix Travis Build Pushed Branches

rsalevsky:


sknorr:
:bouquet: many thanks.

PR #1012, opened 2019-07-10
remove empty procedure step

csymons-suse:

PR #1031, opened 2019-07-12
remove empty step in procedure (no reference, trivial)

csymons-suse:
Remove step-para-/para-/step which created an empty item in this
procedure. Looked odd in the rendered content.

Replaces https://github.com/SUSE-Cloud/doc-cloud/pull/1012


csymons-suse:
Screenshot_empty-step
not worth the effort to get rid of this trivial error

csymons-suse:
seeing is believing

PR #1036, opened 2019-07-15
add hostnamectl to ardana-update-pkgs process (bsc#1138967)

csymons-suse:
the hostnamectl command can be used to demonstrate that
ardana-update-pkgs has made changes. Customer request to
assist with update process.

### DIFF
--- a/xml/operations-maintenance-update_maintenance.xml
+++ b/xml/operations-maintenance-update_maintenance.xml
@@ -165,6 +165,20 @@
         </para>
        </listitem>
       </itemizedlist>
+     </step>
+     <step>
+      <para>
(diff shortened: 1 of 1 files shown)


xml/operations-maintenance-update_maintenance.xml
```diff
@@ -165,6 +165,20 @@
         </para>
        </listitem>
       </itemizedlist>
+     </step>
+     <step>
+      <para>
+       Demonstrate version changes by running <literal>hostnamectl</literal>
```
dmpop: Perhaps _confirm_ instead of _demonstrate_?

csymons-suse: confirm is better. fixed.

PR #1050, opened 2019-07-17
Fix DC file

dmpop:
Change profiling, so it builds properly on Docserv2


sknorr:
Since you need to merge over all of the changes from master -> maintenance/cloud_9 anyway, why not merge everything at once?

sknorr:
Btw, you asked how to fix the Helion 9 content -- I noticed that the DC file for the Helion 9 book is still missing on maintenance/cloud_9. So once that is there, you should be good to go.

asettle:
What @sknorr said but otherwise, looks legitimate ;)

PR #1059, opened 2019-07-18
clarify No Maintenance Mode (bsc#1108818)

csymons-suse:
Crowbar operations guide indicates No Maintenance Mode. This is
misleading. Individual components and services can be put into
maintenance mode.

### DIFF
--- a/xml/soc-operations-maintenance.xml
+++ b/xml/soc-operations-maintenance.xml
@@ -25,14 +25,21 @@
    nodes. Patches that do not require a reboot will not cause a service
    interruption. If a patch (for example, a kernel update) requires a reboot
    after the installation, services running on the machine that is rebooted
-   will not be available within &cloud;.  Therefore it is strongly recommended
-   to install those patches during a maintenance window.
+   will not be available within &cloud;.  Therefore, we strongly recommend
+   installing those patches during a maintenance window.
   </para>
   <note>
-   <title>No Maintenance Mode</title>
+   <title>Maintenance Mode</title>
    <para>
-    As of &productname; 8, it is not possible to put
-    &cloud; into <quote>Maintenance Mode</quote>.
+    As of &productname; 8, it is not possible to put your entire &cloud; into
+    <quote>Maintenance Mode</quote> (such as limiting all users to read-only
+    operations on the control plane), as &ostack; does not support
+    this. However when &pacemaker; is deployed to manage HA clusters, it should
+    be used to place services and cluster nodes into <quote>Maintenance
+    Mode</quote> before performing maintenance functions on them. For more
+    information, see <link
+    xlink:href="https://www.suse.com/documentation/sle_ha/singlehtml/book_sleha/book_sleha.html#cha.ha.configuration.gui">&sle;
+    High Availability documentation</link>.
    </para>
   </note>
   <variablelist>


xml/soc-operations-maintenance.xml
```diff
@@ -31,8 +31,11 @@
   <note>
    <title>No Maintenance Mode</title>
```
aspiers: This title should be changed since there *is* a maintenance mode, but at the Pacemaker level rather than in OpenStack.

xml/soc-operations-maintenance.xml
```diff
@@ -31,8 +31,11 @@
...
    <para>
-    As of &productname; 8, it is not possible to put
-    &cloud; into <quote>Maintenance Mode</quote>.
+    As of &productname; 8, it is not possible to put your entire
+    &cloud; into <quote>Maintenance Mode</quote>, as operations at this level
+    are <literal>read-only</literal>. Services and elements must be put into
```
aspiers: Operations in OpenStack are not read-only so this is not quite right, but I can see where the misunderstanding came from.  In https://bugzilla.suse.com/show_bug.cgi?id=1108818#c0 I wrote:

> My best guess is that it refers to the fact that OpenStack *itself* has no concept of maintenance mode (e.g. allowing only read-only operations on the cloud state / underlying database).

The bit in parentheses was an example of what a maintenance mode in OpenStack *might* provide if it was implemented in the future, not a description of the status quo.

Here is my suggestion for this paragraph:

```xml
It is not possible to put your entire &cloud; into <quote>Maintenance
Mode</quote> (e.g. limiting all users to read-only operations on the
control plane), as OpenStack does not yet support this.  However when
Pacemaker is deployed to manage HA clusters, it should be used to place
the services and cluster nodes it owns into <quote>Maintenance Mode</quote>
before performing maintenance functions on them.  See [INSERT LINK
TO RELEVANT SECTION OF HA GUIDE] for more details on how to do this.
```

xml/soc-operations-maintenance.xml
```diff
@@ -29,13 +29,17 @@
...
-    are <literal>read-only</literal>. Services and elements must be put into
-    <quote>Maintenance Mode</quote> separately to perform maintenance
-    functions.
+    As of &productname; 8, it is not possible to put your entire &cloud; into
+    <quote>Maintenance Mode</quote> (such as limiting all users to read-only
+    operations on the control plane), as &ostack; does not support this.
+    However when &pacemaker; is deployed to manage HA clusters, it should be
+    used to place services and cluster nodes into <quote>Maintenance
+    Mode</quote> before performing maintenance functions on them. For more
+    detail, see <link
```
asettle:
```suggestion
details, see <link
```

xml/soc-operations-maintenance.xml
```diff
@@ -29,10 +29,17 @@
...
    <para>
-    As of &productname; 8, it is not possible to put
-    &cloud; into <quote>Maintenance Mode</quote>.
+    As of &productname; 8, it is not possible to put your entire &cloud; into
+    <quote>Maintenance Mode</quote> (such as limiting all users to read-only
+    operations on the control plane), as &ostack; does not support this.
+    However when &pacemaker; is deployed to manage HA clusters, it should be
+    used to place services and cluster nodes into <quote>Maintenance
+    Mode</quote> before performing maintenance functions on them. For more
+details, see <link
```
aspiers: Ah my bad, missing indentation.

```suggestion
    details, see <link
```

aspiers:
Sorry, looks like the 4 commits need squashing into one and then force-pushing, but I don't have permissions to do that here.

csymons-suse:
squashed, regex-certified. I disagree about "detail" vs "details". It can be "for details" or "for more detail". "For more details" doesn't sound right. Won't change. Seems like a trivial thing for the effort that it created.

csymons-suse:
detail/details bugged me. The more I thought about it the more nonsensical it became. Changed to "information". Please re-re-approve.

asettle:
> detail/details bugged me. The more I thought about it the more nonsensical it became. Changed to "information". Please re-re-approve.

Okay, I'm glad it wasn't just me. I agree that "details" seemed weird, but for some reason I couldn't get behind "detail" either. "information" works perfectly. Cheers.

PR #1072, opened 2019-07-23
add image-volume cache instructions for SES (bsc#1140663)

csymons-suse:
image-volume caches must be enabled in configuration


csymons-suse:
needs review from Andreas Mach

PR #1092, opened 2019-08-01
SOC-8980: Moving new sections in CLM install guide

asettle:
The three sections do not fall under UI verifications. Moving to post-installation section where the other verification section is.


asettle:
> would be interesting to know how this was misplaced. Good catch.

I agree. The bug report indicates it was picked up by QE during their UI verification stage in April.

asettle:
I know Dima's away for tomorrow, moving forward with this.

PR #1095, opened 2019-08-01
add ardana prompt, add ardana-init step (bsc#1143310)

csymons-suse:
Commands need to be run as user ardana.
User ardana is created with ardana-init.

### DIFF
--- a/xml/installation-preparing-standalone.xml
+++ b/xml/installation-preparing-standalone.xml
@@ -116,7 +116,7 @@
  </itemizedlist>
  </section>
  <section>
-  <title>Configuring a Stand-Alone Deployer</title>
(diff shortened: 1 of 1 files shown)


rtamalin:
FYI the ardana user account is created, if it doesn't already exist, as a result of installing the base `ardana` package which happens as a result of installing the Cloud pattern, not when the `ardana-init` command is run... Aspects of the ardana user account will be setup/updated by running ardana-init, but a skeletal account will already have been created, if needed, by the `ardana` package being installed

csymons-suse:
The content added at line 137 is the same as http://docserv.nue.suse.com/documents/SUSE_OpenStack_Cloud_9/suse-openstack-cloud-deployment/single-html/ 21.1 Prepare for Cloud Installation

PR #1091, opened 2019-08-01
SOC-8946: Update Ovsvapp install docs

asettle:
Two of three items within the bug have been updated, one more requires further clarification.

PR #1103, opened 2019-08-06
Update Ovsvapp install docs (SOC-8946)

asettle:
Two of three items within the bug have been updated, one
more requires clarification.

### DIFF
--- a/xml/config-ansible_playbook.xml
+++ b/xml/config-ansible_playbook.xml
@@ -44,14 +44,17 @@
    </para>
   </listitem>
  </itemizedlist>
- <para>
-  In order to execute the playbook from the &clm;, the
-  <literal>python-networking-vsphere</literal> package must be installed.
- </para>
- <screen>&prompt.sudo;zypper install python-networking-vsphere</screen>
- <para>
-  Run the playbook:
- </para>
+<para>The <filename>neutron-create-ovsvapp-resources.yml</filename> playbook is
+  not set up by default. In order to execute the playbook from the &clm;, the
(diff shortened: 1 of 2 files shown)


asettle:
Related to: https://github.com/SUSE-Cloud/doc-cloud/pull/1091

xml/config-ansible_playbook.xml
```diff
@@ -44,25 +44,28 @@
...
+ <step>
+   <para>Run the playbook:</para>
  <screen>&prompt.ardana;ansible-playbook neutron-create-ovsvapp-resources.yml \
 -i hosts/verb_hosts -vvv -e 'variable_host=localhost
 vcenter_username=<replaceable>USERNAME</replaceable>
 vcenter_encrypted_password=<replaceable>ENCRYPTED_PASSWORD</replaceable>
 vcenter_ip=<replaceable>IP_ADDRESS</replaceable>
 vcenter_port=443
 vc_net_resources_location=<replaceable>LOCATION_TO_RESOURCE_DEFINITION_FILE</replaceable>
-'
```
csymons-suse: The lonely single quote on line 62 is puzzling. It looks odd to have a single quote (maybe an opening quote) in line 56/59 with no (possible) closing quote. 

Perhaps `variable_host=localhost` belongs on a new line as the first line of the playbook output? Or is `variable_host=localhost` part of the playbook command? I dunno.
Are the quotes part of the output? Are they part of the command?

asettle: In the read == removed. I accidentally added it (rushing and pressing buttons?) so it's gone now. But yes, totally understandable concern but totally gone.

PR #1117, opened 2019-08-12
MANAGEMENT network group cannot be changed, is required (SOC-10106)

csymons-suse:
MANAGEMENT network group name is not just a good idea. Must be in
upper case, cannot be changed.


jeremy-moffitt:
@rktidwell should we be including this note in Cloud 8? I'm assuming only Cloud 9 for now, with the alternate information from https://jira.suse.com/browse/SOC-10150 going into the Cloud 8 docs

PR #1120, opened 2019-08-13
add workaround for partition image resize (bsc#1145498)

csymons-suse:
deployment documentation covers ISO image size exceeding
free space. Add instructions for increasing /tmp size for
partition images.


csymons-suse:
rendered version:
Screen_partition-image-resize

PR #1124, opened 2019-08-19
add requirement for dummy entries in servers.yml (bsc#1146206)

csymons-suse:
when servers are imaged with user-provided tooling, dummy entries
must be entered in servers.yml to prevent cobbler-deploy.yml failure


csymons-suse:
content added to multiple locations as appropriate to deployment/installation guides in master, cloud 9 and cloud 8. Also added to Operations guides where needed.
Screenshot_dummy-password

PR #1133, opened 2019-08-21
instruction for bootstrapping after MariaDB update (bsc#1132852)

csymons-suse:
the compute plane needs to be rebooted after MariaDB update.
Instructions aren't obvious.

### DIFF
--- a/xml/installation-kvm_xpointer.xml
+++ b/xml/installation-kvm_xpointer.xml
@@ -540,6 +540,17 @@
      installing the &clm; from the ISO, and is the same user that is running
      the cobbler-deploy play.
     </para>
+    <note>
+     <para>
+      When imaging servers with your own tooling, it is still necessary to have
+      ILO/IPMI settings for all nodes. Even if you are not using Cobbler, the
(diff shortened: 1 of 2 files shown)


xml/depl_maintenance.xml
```diff
@@ -2174,6 +2174,12 @@ migrate</screen>
        Or take the cluster out of maintenance mode.
       </para>
      </step>
+     <step>
+      <para>
+       If necessary, boot the compute plane with the instructions in <xref
+       linkend="sec-depl-maintenance-bootstrap-compute-plane"/>.
```
rhafer: I think this is targetting the wrong section. AFAIU the idea is to reference the section about recovering galera somehow: https://github.com/SUSE-Cloud/doc-cloud/blob/maintenance/cloud_9/xml/soc-operations-maintenance.xml#L1338 but obviously that is not available in Cloud 8 (yet?).

However adding that note about that here is also misleading I think. Because the step outline above won't really work if the database cluster is down.

How about just adding a note at the beginning of the chapter, that in order to run these update steps the database cluster needs to be up and healthy.

PR #1134, opened 2019-08-21
clarify domain and project admin roles (bsc#1145749)

csymons-suse:
admin roles need to be assigned by domain and by project
to ensure the ability to add users to projects and show
Owner information in Horizon Compute-Images panel


csymons-suse:
Screenshot_domain-admin

PR #1140, opened 2019-08-23
Create LICENSE for CC-BY-SA (no-ref)

asettle:
I don't know how, but we haven't had a license for
the repo yet. This patch is for the CC one, if
we choose to go with it.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,319 @@
+Creative Commons Legal Code
+
+Attribution 3.0 Unported
+
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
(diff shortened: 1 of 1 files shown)


asettle:
"This is a Pull Request, therefore will not build." - that's a new one

sknorr:
@asettle In your commit message, you need to use `(noref)` instead of `(no-ref)`.

Also, this is a CC-BY-SA 3.0 license whereas your docs currently reference "CC-BY 3.0" (http://creativecommons.org/licenses/by/3.0/legalcode/).

sknorr:
There is a nicely-formatted plaintext version of the license at:
  https://creativecommons.org/licenses/by/3.0/legalcode.txt

Also, the in-document link apparently needs to be changed to `https://creativecommons.org/licenses/by/3.0/legalcode` (note the httpS) -- the current URL just redirects to the homepage.

PR #1144, opened 2019-08-23
Update README with license info (noref)

asettle:
Adds additional ref to licensing


asettle:
Need further clarification on the licensing from Ciaran and Christopher regarding the Crowbar ops guide.

asettle:
Christopher is reviewing all of the repos and packaging to determine what's the best option.

PR #1160, opened 2019-09-16
Swift doc cleanup (SOC-8649)

dmpop:
Add and update info on adding and removing disks.

### DIFF
--- a/xml/operations-objectstorage-input_model_change_existing_rings.xml
+++ b/xml/operations-objectstorage-input_model_change_existing_rings.xml
@@ -282,7 +282,7 @@ the required actions. More detailed procedures are provided elsewhere.-->
           </listitem>
           <listitem>
            <para>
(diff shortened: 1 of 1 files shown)


xml/operations-objectstorage-input_model_change_existing_rings.xml
```diff
@@ -416,6 +435,20 @@ the required actions. More detailed procedures are provided elsewhere.-->
...
+          <row>
+      <entry>
+       Removing Disk Drive(s)
+      </entry>
+      <entry>
+       <para>
+	When removing a disk drive from the input model, keep in mind that this
+	drops the disk out of the ring without allowing Swift to move the data
+	off it first. While it should be fine in a properly replicated healthy
+cluster, we do not recommend this approach. A better solution is to step
```
asettle:
```suggestion
    cluster, we do not recommend this approach. A better solution is to step
```

PR #1169, opened 2019-09-20
Adding PCI Passthrough section to ops crow (SOC-10568)

asettle:
This also adds a new section, managing networking for
future updates.

### DIFF
--- a/xml/soc-operations-gpu-passthrough.xml
+++ b/xml/soc-operations-gpu-passthrough.xml
@@ -131,13 +131,13 @@ drwxr-xr-x 3 root root 0 Feb 19 16:09 /sys/kernel/iommu_groups/20/devices/0000:0
          We suggest user to create new <filename>nova.conf</filename> file
          specific to pci passthrough usage under
          <literal>/etc/nova/nova.conf.d</literal> instead of modifying the
-         existing <filename>nova.conf</filename> to avoid any side effect
-         causing from chef-client.service. For example,
+         existing <filename>nova.conf</filename> to avoid any side effects
+         caused from the <literal>chef-client.service</literal>. For example,
          <filename>100-nova.conf</filename> or
-         <filename>101-nova-placement.conf</filename>. Please note that this
-         new <filename>nova.conf</filename> will not be recorded in Nova
-         proposal and under manipulation by chef-client.service. Please make
-         sure you have an backup for that just in case for future incident.
+         <filename>101-nova-placement.conf</filename>. Note that this
+         new <filename>nova.conf</filename> will not be recorded in any nova
+         proposal and under manipulation by the <literal>chef-client.service</literal>.
+         Ensure you have backups in case of any future incidents.
        </para>
      </note>
    </listitem>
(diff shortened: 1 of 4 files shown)


asettle:
Requires further review from Kent Wu and potentially Meera Belur.

xml/soc-operations-networking.xml
```diff
@@ -0,0 +1,166 @@
...
+          <para>
+            If you add them on the existing <literal>nova.conf</literal>, then it
+            is overwritten by the <literal>chef-client.service</literal> and
+            restored to the original version of <literal>nova.conf</literal>.
+          </para>
+        </note>
+      </step>
+      <step>
+        <para>
+          Add the following configuration to <literal>nova.conf</literal> on
```
kwu83tw: If we choose to not duplicate the effort/content, then the only difference here is we want user to create new nova configuration file instead of editing the existing one by following the openstack doc.

Here is just showing an example of what you may find the nova.conf in your &products environment. Then we ask user to create new nova.conf file under /etc/nova/nova.conf.d instead of modifying the existing one, e.g, 100-nova.conf, 101-nova-placement.conf

May you suggest a better way to handle this so that the customer can understand that?

xml/soc-operations-networking.xml
```diff
@@ -0,0 +1,166 @@
...
+          the controller node:
+        </para>
+        <screen>
+          /etc/nova/nova.conf.d # ls
+          010-nova.conf  100-nova.conf  101-nova-placement.conf
+        </screen>
+      </step>
+      <step>
+        <para>
+          Add filter <literal>PciPassthroughFilter</literal> to <literal>enabled_filters</literal>.
```
kwu83tw: As for this paragraph for adding PciPassthroughFilter, we propose a [PR](https://github.com/crowbar/crowbar-openstack/pull/2247) to set it as a default value when user choose all filters. I think we can ignore this paragraph for now.

xml/soc-operations-networking.xml
```diff
@@ -0,0 +1,166 @@
...
+    </procedure>
+   </listitem>
+   <listitem>
+    <para>
+     <emphasis role="bold">Preparing &o_comp; and &o_img; for
+     passthrough</emphasis>
+    </para>
+    <procedure>
+      <step>
+        <para>
```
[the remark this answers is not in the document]
asettle: Will update the incorrect link, thanks.

Regarding the rest, I agree, let's not rewrite.

xml/soc-operations-gpu-passthrough.xml
```diff
@@ -0,0 +1,146 @@
...
+    <para>
+     <emphasis role="bold">Preparing &o_comp; and &o_img; for
+     passthrough</emphasis>
+    </para>
+    <para>
+      Information about configuring &o_comp; is available in the
+      documentation at
+      <link
+      xlink:href="https://docs.openstack.org/nova/rocky/admin/pci-passthrough.html"/>.
+      Services like <literal>nova-compute</literal>, <literal>nova-scheduler</literal>
```
kwu83tw: I don't see any glance related thing in the doc. So I remove it.

xml/soc-operations-gpu-passthrough.xml
```diff
@@ -0,0 +1,146 @@
...
+      and <literal>nova-api</literal> must be configured.
+     </para>
+     <note>
+       <para>
+         We suggest user to create new <filename>nova.conf</filename> file
```
kwu83tw: Add a bit more context here to help customer know the pros and cons of creating unversioned file.

PR #1191, opened 2019-10-01
message to add non-admin node for public network (SOC-10658)

csymons-suse:
admin node is not added to public network. Yet at least one node
is necessary that can be added to the public network.


csymons-suse:
> Looks good Carl!

Thanks Ritesh

PR #1209, opened 2019-10-31
master: Use GeekoDoc again

sknorr:
This branch fixes the minor issues that made the documents incompatible with GeekoDoc and it removes the DC file settings that made us always validate with DocBook instead of GeekoDoc.

To make sure you validate with GeekoDoc locally:

1. install the `geekodoc` package
2. Set DAPS to use GeekoDoc by default: `echo "DOCBOOK5_RNG_URI=https://github.com/openSUSE/geekodoc/raw/master/geekodoc/rng/geekodoc5-flat.rnc" > ~/.config/daps/dapsrc`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,7 @@ matrix:
         - echo "Building docker image"
         - docker build -t sle-doc-image .
       script:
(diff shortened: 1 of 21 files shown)


sknorr:
And sorry, Alex, couldn't help it.

PR #1218, opened 2019-12-02
Clarify VSA/Ceph support in HOS 8 , SOC-10981 (bsc#1144694)

jeremy-moffitt:
Clarify VSA support for HOS8, emphasize that customers need to migrate to external SES clusters before migrating.

### DIFF
--- a/xml/art_migration.xml
+++ b/xml/art_migration.xml
@@ -114,6 +114,14 @@
    </listitem>
    <listitem>
     <para>
+     &productname; &productnumber; dropped support for internally deployed
+     storage such as Ceph and VSA. Data must be moved out of existing
+     internal storage backends to an external SES cluster before performing
+     an upgrade.
+    </para>
+   </listitem>
+   <listitem>
+    <para>
      The cloud data model does not contain services which are not included in
      &productname; &productnumber;, such as Ceph and VSA.
     </para>


jeremy-moffitt:
soc-10981

xml/art_migration.xml
```diff
@@ -112,6 +112,13 @@
      Guide and Operations Guide.
     </para>
    </listitem>
+   <listitem>
+    <para>
+     &productname; &productnumber; dropped support for internally deployed
+     storage such as Ceph and VSA, move data out of the old internal storage
```
asettle: This bit doesn't make any sense. Do you mean to say it dropped support BY moving data out of the old internal storage? I'm not sure if this is supposed to be two separate sentences.

KeithMnemonic: It means that prior to the migration, customers running HOS need to remove any VSA or  "non SES" Ceph backends deployed by HOS. The data in these should be migrated to a supported backend such as 3PAR or SES. 

The reason for this was customers who may have deployed a newer ceph outside of what was provided by HOS.

jeremy-moffitt: I tweaked the wording of this a bit, please let me know if it makes more (or less) sense now.

PR #1220, opened 2019-12-03
Clarify keyring chown instructions for Ceph (bsc#1111180)

jeremy-moffitt:
Per https://bugzilla.suse.com/show_bug.cgi?id=1111180 the documentation at https://documentation.suse.com/soc/8/single-html/suse-openstack-cloud-deployment/#sec-depl-inst-nodes-post-ceph-ext-install  refers to a non-existent file /etc/ceph/ceph.volumes.keyring  in step 6.d , additionally the way that step is organized appears tacked on and is a bit confusing. 

Moved chown commands up into step 6.c , splitting by service type, for the sake of clarity.


jeremy-moffitt:
bz1111180

PR #1221, opened 2019-12-03
Update incorrect Manila install/setup instructions (SOC-10975)

jeremy-moffitt:
Per https://bugzilla.suse.com/show_bug.cgi?id=1149860  (also https://jira.suse.com/browse/SOC-10975 ), the Cloud 8 documentation for Manila is wrong. Using the Cloud 9 documentation (which is correct) is the present workaround. This change copies the Cloud 9 Manila docs into Cloud 8. The only change from the Cloud 9 content is a replacement of &product; with &productname; as the former does not exist in Cloud 8.

### DIFF
--- a/xml/installation-ardana-manila.xml
+++ b/xml/installation-ardana-manila.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!DOCTYPE chapter [
- <!ENTITY % entities SYSTEM "entities.ent"> %entities;
+ <!ENTITY % entities SYSTEM "entity-decl.ent"> %entities;
 ]>
 <chapter xml:id="install-ardana-manila"
  xmlns="http://docbook.org/ns/docbook" version="5.1"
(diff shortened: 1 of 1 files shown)


xml/installation-ardana-manila.xml
```diff
@@ -176,11 +441,11 @@ on &clm;:
 --protocol icmp
 &prompt.ardana;openstack security group rule list manila-security-group (verify manila security group)
 &prompt.ardana;openstack keypair create --public-key ~/.ssh/id_rsa.pub mykey
-&prompt.ardana;neutron net-create n1
-&prompt.ardana;neutron subnet-create --name s1 n1 <replaceable>11.11.11.0/24</replaceable>
-&prompt.ardana;neutron router-create r1
-&prompt.ardana;neutron router-interface-add r1 s1
-&prompt.ardana;neutron router-gateway-set r1 ext-net
+&prompt.ardana;openstack network create n1
+&prompt.ardana;openstack subnet create --name s1 n1 <replaceable>11.11.11.0/24</replaceable>
```
nicolasbock: `openstack subnet create s1 --network n1 <replaceable>--subnet-range 11.11.11.0/24</replaceable>`

xml/installation-ardana-manila.xml
```diff
@@ -273,7 +539,7 @@ default_share_type = default1</screen>
      Create a &o_sharefs; share image and verify it
     </para>
 <screen>&prompt.ardana;wget http://tarballs.openstack.org/manila-image-elements/images/manila-service-image-master.qcow2
-&prompt.ardana;. service.osrc;glance image-create --name <literal>manila-service-image-new</literal> \
+&prompt.ardana;. service.osrc;openstack image create --name <literal>manila-service-image-new</literal> \
```
nicolasbock: `openstack image create <literal>manila-service-image-new</literal>`

xml/installation-ardana-manila.xml
jeremy-moffitt: updated, does this mean that the Cloud 9 docs need to be updated as well, or was this changed between Pike and Rocky?

xml/installation-ardana-manila.xml
jeremy-moffitt: updated, same question as above... does this mean that the Cloud 9 docs need to be updated as well, or was this changed between Pike and Rocky?

xml/installation-ardana-manila.xml
nicolasbock: I don't think it changed. It's more a matter of whether you use the `neutron` CLI or the `openstack` client. If we use the `neutron` client in the Cloud 9 docs we should update those as well. Whenever possible we should be advocating the `openstack` client.

xml/installation-ardana-manila.xml
nicolasbock: ditto

PR #1224, opened 2019-12-05
Fix extraneous --name parameter (SOC-10975)

jeremy-moffitt:
In reviewing these docs for SOC8, feedback was received
that the --name parameter is unnecessary and may be a legacy
from before the "openstack" cli was used. Removed it here


jeremy-moffitt:
based on feedback from https://github.com/SUSE-Cloud/doc-cloud/pull/1221 , the cloud 9 manila docs need a slight tweak, making this change on master and then will cherry-pick it to maintenance/cloud_9

PR #1227, opened 2019-12-06
Fix extraneous --name parameter (SOC-10975)

jeremy-moffitt:
In reviewing these docs for SOC8, feedback was received
that the --name parameter is unnecessary and may be a legacy
from before the "openstack" cli was used. Removed it here

cherry-pick e6f60ac605bc12c62130483ec6a8aeb5eb7f14bb


asettle:
Merging asap as it's a cherry pick.

PR #1229, opened 2019-12-10
Remove deprecated manila-share instruction step from Cloud 9 doc (SOC-10938)

jeremy-moffitt:
manila-share 's are installed on the same hosts as manila-api in Cloud 9, remove the extraneous instruction


jeremy-moffitt:
soc-10938-cloud9

PR #1230, opened 2019-12-10
Specify that manila-share be installed on the control node (SOC-10938)

jeremy-moffitt:
Instructions currently incorrectly state that manila-share should be
installed on compute nodes. Updated to correctly specify control
node

note that this is different from the change at
https://github.com/SUSE-Cloud/doc-cloud/pull/1229
because SOC 9 ignores the manila-share config and installs
it where manila-api is installed


jeremy-moffitt:
soc-10938-cloud8

PR #1234, opened 2019-12-11
Remove deprecated manila-share instruction from Cloud 9 doc (SOC-10938)

jeremy-moffitt:
manila-share is deployed on the same hosts as manila-api
in Cloud 9, and should not have a separate config instruction


jeremy-moffitt:
cherry-pick from master

PR #1235, opened 2019-12-11
Travis/Gitlint: switch config file, allow body-less commits (noref)

sknorr:
As discussed among Jeremy, Dima, Alex, me, relax the rules for the gitlint commit message checker in Travis a little. This switches us to a different config file, provided from a doc-team repo (which I hope is fine). We now also have better error messages if there is an issue with the title line's bug reference (i.e. no more regexes thrown in people's face).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ matrix:
       if: type = pull_request
       script:
       - pip install gitlint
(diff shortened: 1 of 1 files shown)


sknorr:
To be fair, @jeremy-moffitt, we don't use it in any other doc repo (yet) but changing the SOC file we used previously would change behavior in other SOC-related repos that are also using the existing config file.

sknorr:
@nicolasbock Done.

PR #1242, opened 2020-01-15
Unsupported upgrade configs

dmpop:
Add info about unsupported configurations as per https://jira.suse.com/browse/SOC-10496

### DIFF
--- a/xml/soc-operations-maintenance.xml
+++ b/xml/soc-operations-maintenance.xml
@@ -544,7 +544,7 @@
      <itemizedlist>
     <listitem>
      <para>
-      Run <command>nova-service disabled</command> on all &o_comp; &compnode;s
+      Run <command>nova service-disable</command> for all &o_comp; &compnode;s
       running the Xen hypervisor. In the Crowbar context, these are the
       &compnode;s with the <literal>nova-compute-xen</literal> role in the
       &o_comp; barclamp. This prevents any new VMs from being launched on these
--- a/xml/soc-operations-maintenance.xml
+++ b/xml/soc-operations-maintenance.xml
@@ -491,7 +491,7 @@
     prechecks fail if they detect an unsupported configuration. Below is a list
     configurations that cause failure and possible solutions.
    </para>
-   <variablelist><title>Unsupported configurations</title>
+   <variablelist><title>Unsupported Configurations</title>
   <varlistentry>
     <term>
       &aodh; deployed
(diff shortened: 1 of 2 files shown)


xml/soc-operations-maintenance.xml
```diff
@@ -482,6 +482,107 @@
     </listitem>
    </itemizedlist>
   </section>
+
+  <section xml:id="sec-upgrade-unsupported-configs">
+   <title>Unsupported configurations</title>
+   <para>
+    In &cloud; 9, certain configurations and barclamp combinations are no
+    longer supported. See the &cloud; 9 release notes for details. The upgrade
```
asettle: Can we link to the release notes?

xml/soc-operations-maintenance.xml
```diff
@@ -482,6 +482,107 @@
...
+    <listitem>
+     <para>
+       As of &cloud; 9, the Xen hypervisor is no longer supported. On any
+       cloud, where Xen based compute nodes are still present, the following
+       procedure must be followed before it can be upgraded to &cloud; 9:
+     </para>
+     <itemizedlist>
+    <listitem>
+     <para>
+      Run <command>nova-service disabled</command> on all &o_comp; &compnode;s
```
jgrassler: Yikes. I really messed up that part of the document (I checked, the errors are present in my asciidoc text already) now that I look at it again in a different format. There are two things wrong with this:

1. `nova service-disable` (without the `d` at the end and with the dash moved)
2. Run `nova service-disable` _for_ all compute nodes - that command is being run on a controller for every compute node to be disabled.

dmpop: No problem whatsoever! Fixed. :-)

xml/soc-operations-maintenance.xml
```diff
@@ -482,6 +482,109 @@
...
    </itemizedlist>
   </section>
+
+  <section xml:id="sec-upgrade-unsupported-configs">
+   <title>Unsupported configurations</title>
+   <para>
+    In &cloud; 9, certain configurations and barclamp combinations are no
+    longer supported. See the <link
+    xlink:href="https://www.suse.com/releasenotes/x86_64/SUSE-OPENSTACK-CLOUD/9/">&cloud;
+    9 release notes</link> for details. The upgrade prechecks fail if they
```
jsuchome: The pdf shows the URL as visible part of the text, is that correct?

dmpop: Yes, PDF show the full link (useful when you print docs).

PR #1253, opened 2020-01-29
Add instructions for lbaas v2 loadbalancers (SOC-10980)

jeremy-moffitt:
Instructions on how to delete an errored loadbalancer created with
Neutron's lbaas v2 API.


jeremy-moffitt:
cherry-picked to the maintenance/cloud_8 branch, though not 100% sure that the instructions should be the same for SOC8 as SOC9, need confirmation from the reviewers if possible

PR #1254, opened 2020-02-03
Refactor Octavia maintenance (SOC-8726)

nicolasbock:
This change pulls the Octavia maintenance sections out into a separate
file and links that file from the CLM and Crowbar Operations manuals.

Signed-off-by: Nicolas Bock <nicolas.bock@suse.com>


nicolasbock:
I haven't figured out yet how to keep the prompts consistent when building the Ardana documentation as opposed to the Crowbar documentation. I'll look whether one can define the prompt macro depending on the life cycle manager.

jeremy-moffitt:
I dont think the prompts are too big an issue, looking at the single page html, `tux` is used in almost 300 places... so that may be ok. I'm surprised at the scope of changes to `soc-operations-maintenance.xml`  is that intentional?

nicolasbock:
I changed a lot of indentations. I don't know if we have a "coding style" for XML, but the indentation was pretty inconsistent.

nicolasbock:
I can revert that of course.

jeremy-moffitt:
If the content is the same (other than embedding the removal of load balancers) I wont scan it for content, but just sanity check spacing, etc... thanks for the info. The CLM docs look fine, building Crowbar next.

nicolasbock:
No content changes

PR #1255, opened 2020-02-04
Designate: recommend deploying DNS in a cluster in HA deployment (SOC-10636)

guangyee:
…10636)

In HA deployment, we should recommend users to deploy both DNS and Designate
into the same cluster nodes to avoid single point of failure.

### DIFF
--- a/xml/depl_nodes.xml
+++ b/xml/depl_nodes.xml
@@ -171,6 +171,12 @@
     barclamp that is not the admin node. The admin node is not added to the
     public network. So another node is needed that can be attached to the public
     network and appear in the designate default pool.
+
+    In High Availability deployments where Designate services are running in a cluster,
+    it is highly recommended that DNS services are running in a cluster as well.
+    For example, in a typical High Availability deployment where the controllers
+    are deployed in a 3-node cluster, the DNS barclamp should be applied to all the
+    controllers, in the same manner as Designate.
    </para>
   </note>
   <variablelist>


xml/depl_nodes.xml
```diff
@@ -171,6 +171,11 @@
     barclamp that is not the admin node. The admin node is not added to the
     public network. So another node is needed that can be attached to the public
     network and appear in the designate default pool.
+
+    In High Availability deployment where Designate services are running in a cluster,
+    it is highly recommended that DNS services are running in the same cluster as well.
+    Therefore, DNS barclamp should be applied the the same nodes as the
```
nicolasbock:
```suggestion
    Therefore, the DNS barclamp should be applied to the same nodes as the
```

nicolasbock: Why is it better to run DNS in the same cluster? I get that a DNS cluster is a good thing (:tm:) but what's the advantage of it being the same cluster?

guangyee: You're right. It doesn't have to be in the same cluster. But we usually group all the related services in the same cluster right? I can change it from "same cluster" to "a cluster". Please let me know.

nicolasbock: Ah ok, that makes sense. Yes, in CI we do, but who is to say what a customer might prefer? :wink: Let's just change it to "a cluster" as you suggest. Then we are definitely on the safe side :smile:

guangyee: done

xml/depl_nodes.xml
```diff
@@ -171,6 +171,12 @@
     barclamp that is not the admin node. The admin node is not added to the
     public network. So another node is needed that can be attached to the public
     network and appear in the designate default pool.
+
+    In High Availability deployment where Designate services are running in a cluster,
```
nicolasbock:
```suggestion
    In High Availability deployments where Designate services are running in a cluster,
```

xml/depl_nodes.xml
```diff
@@ -171,6 +171,12 @@
...
+    it is highly recommended that DNS services are running in a cluster as well.
+    For example, in a typical High Availability deployment where the controllers
+    are deployed in 3-node cluster, DNS barclamp should be applied to all the
```
nicolasbock:
```suggestion
    are deployed in a 3-node cluster, the DNS barclamp should be applied to all the
```

xml/depl_nodes.xml
guangyee: done

xml/depl_nodes.xml
guangyee: good catch!

PR #1258, opened 2020-03-05
Updating SES integration documentation (JSC#SOC-11140)

asettle:


xml/depl_nodes.xml
```diff
@@ -7640,6 +7640,28 @@ radosgw_urls:
        No special configuration is needed for &o_comp;. It uses the
        configuration from the &o_blockstore; &barcl; to connect with volumes
        stored in SES.
+       For ephemeral storage, &o_comp; re-uses the <literal>rbd_store_user</literal>
+       and key from &o_blockstore; but has a separate <literal>rbd_store_pool</literal>
```
skazi0: `key` should also be a literal

xml/depl_nodes.xml
```diff
@@ -7640,6 +7640,28 @@ radosgw_urls:
...
+       defined in the SES configuration.
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term>RadosGW Integration</term>
```
skazi0: Ardana docs use "RADOS Gateway" term here. Maybe we should do the same.

xml/depl_nodes.xml
```diff
@@ -7640,6 +7640,28 @@ radosgw_urls:
...
+     <listitem>
+      <para>
+       Besides &o_blockstore;, SES cluster can also be used as a &o_objstore;
```
skazi0: &o_blockstore; will translate to Cinder, right? If so, I think it's better to say "block storage" which is a bit more abstract and matches "object storage" at the end of this sentence.

xml/depl_nodes.xml
```diff
@@ -7640,6 +7640,28 @@ radosgw_urls:
...
+       replacement for object storage. If <literal>radosgw_urls</literal> section is present
+       in uploaded SES configuration, first of the URLs are registered
```
skazi0: "is registered"?

xml/depl_nodes.xml
asettle: https://github.com/SUSE-Cloud/doc-cloud/blob/maintenance/cloud_9/xml/entity-decl.ent#L270

Does indeed. At the end it translates to swift: https://github.com/SUSE-Cloud/doc-cloud/blob/maintenance/cloud_9/xml/entity-decl.ent#L95 

So it currently does match, but if you'd prefer the change I can do that.

xml/depl_nodes.xml
```diff
@@ -7640,6 +7640,28 @@ radosgw_urls:
...
+       in the &o_ident; catalog as "&swift;"/"object-store" service. Some
```
skazi0: what's the difference between `&swift;` and `&o_objstore;`?

asettle: You can see all of the entitys here: https://github.com/SUSE-Cloud/doc-cloud/blob/maintenance/cloud_9/xml/entity-decl.ent#L76

xml/installation-installation-ses_integration.xml
```diff
@@ -63,9 +63,9 @@ glance:
   rbd_store_user: glance
 nova:
   rbd_store_pool: cloud-vms
-  radosgw_urls:
-     - http://10.84.56.7:80/swift/v1
-     - http://10.84.56.8:80/swift/v1 </screen>
+radosgw_urls:
+  - http://10.84.56.7:80/swift/v1
+  - http://10.84.56.8:80/swift/v1 </screen>
```
skazi0: the second example has this list indented too much too.

xml/depl_nodes.xml
skazi0: I meant the literal "object storage" we have at the end of the sentence on following line.

xml/depl_nodes.xml
skazi0: thx!

xml/depl_nodes.xml
asettle: I see. I can change that.

jeremy-moffitt:
superseded by https://github.com/SUSE-Cloud/doc-cloud/pull/1259

PR #1270, opened 2020-04-16
Octavia: update documentation on lb-mgmt-net (SOC-10904)

stefannica:
Update the documentation to cover the new supported way of configuring
the Octavia management network through an `octavia` network barclamp
configuration entry.

### DIFF
--- a/xml/depl_nodes.xml
+++ b/xml/depl_nodes.xml
@@ -6837,8 +6837,7 @@ monitoring</guimenu>
     <sect3 xml:id="sec-depl-ostack-octavia-mgmtnet">
       <title>Management network</title>
       <para>
-        Octavia uses a load balancer network,
-        <literal>lb-mgmt-net</literal>, as a management network that the
+        Octavia needs a neutron provider network as a management network that the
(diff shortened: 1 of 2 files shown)


xml/depl_nodes.xml
```diff
@@ -6847,37 +6846,69 @@ monitoring</guimenu>
...
+              "broadcast": "<replaceable>172.31.255.255</replaceable>",
+              "ranges": {
+                "host": { "start": "<replaceable>172.31.0.1</replaceable>",
+                   "end": "<replaceable>172.31.0.255</replaceable>" },
+                "dhcp": { "start": "<replaceable>172.31.1.1</replaceable>",
+                   "end": "<replaceable>172.31.255.254</replaceable>" }
+              }
+        },</screen>
+         <para>
+          Care should be taken to ensure the IP subnet doesn't overlap with any of those
```
guangyee, line 6880: Make sure vlan is unique as well?

stefannica: @guangyee oh, right, that also reminds me I need to warn them about using a VLAN ID that is assigned to the fixed network

PR #1289, opened 2020-09-04
Clarify reboot instructions during update workflow (SOC-11386)

jeremy-moffitt:
specified that the reboot should only take place after the update playbook has been run. 

clmupdate_update


KeithMnemonic (approved):
LGTM

PR #1294, opened 2020-09-15
removed all nova-consoleauth entries for soc9 documentation (SOC-11356)

yarunachalam:
Remove nova-consoleauth entries for all soc9 documentation as per soc-11356


guangyee (approved):
Looks good to me.

jeremy-moffitt (approved):
LGTM! I compared to the live version and looks like you got all of the places :)

PR #1295, opened 2020-09-16
Soc 11356crow

yarunachalam:
Removed nova-consoleauth entry from deploy nodes list from crowbar documentation as per - soc-11356 ticket.


jeremy-moffitt:
This should be rebased, also need to figure out why the table is unhappy in depl_nodes with this change

jeremy-moffitt:
pushed a rebase

jeremy-moffitt (approved):
LGTM

PR #1296, opened 2020-09-21
Soc 11000

yarunachalam:
Replaced PostgreSQL to MariaDB as per comment4&5 in Bugzila


jeremy-moffitt:
think this was meant for maintenance/cloud_8  , closing

PR #1298, opened 2020-10-06
Modified the PTF install instructions as per the (SOC-10616)

yarunachalam:
Validated the PTF install instructions and modified as per the comments.


jeremy-moffitt:
soc-10616_0
screenshot of the modified version

rtamalin:
The fixed step 9 looks good above... But possibly step 8 and 9 above should be reverse so that we install the venv on the deployer before doing any actual updating of the nodes...

rtamalin (approved):
LGTM

jeremy-moffitt (approved):
LGTM

PR #1301, opened 2020-10-07
added database maintenance section (SOC-9876)

yarunachalam:
Added database maintenance section

### DIFF
--- a/xml/operations-maintenance-database_maintenance.xml
+++ b/xml/operations-maintenance-database_maintenance.xml
@@ -6,7 +6,7 @@
  <!ENTITY % entities SYSTEM "entity-decl.ent"> %entities;
 ]>
 <section xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude"
- xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="planned-maintenance">
+ xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="database-maintenance">
  <title>Database Maintenance</title>
   <para>
   As part of Database maintenance for your cloud. See sections below for cleanup commands:
--- a/xml/operations-maintenance-database_maintenance.xml
+++ b/xml/operations-maintenance-database_maintenance.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0"?>
+<?xml-stylesheet href="urn:x-suse:xslt:profiling:docbook51-profile.xsl"
+ type="text/xml"
+ title="Profiling step"?>
+<!DOCTYPE section [
+ <!ENTITY % entities SYSTEM "entity-decl.ent"> %entities;
+]>
+<section xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude"
+ xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="planned-maintenance">
+ <title>Database Maintenance</title>
+  <para>
+  As part of Database maintenance for your cloud. See sections below for cleanup commands:
+ </para>
+<screen>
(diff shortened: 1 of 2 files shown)


xml/operations-maintenance-database_maintenance.xml
```diff
@@ -0,0 +1,29 @@
...
+ type="text/xml"
+ title="Profiling step"?>
+<!DOCTYPE section [
+ <!ENTITY % entities SYSTEM "entity-decl.ent"> %entities;
+]>
+<section xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude"
+ xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="database-maintenance">
+ <title>Database Maintenance</title>
+  <para>
+  As part of Database maintenance for your cloud. See sections below for cleanup commands:
```
jeremy-moffitt: I would remove the word "sections" and just have this read "See below for cleanup commands:"

xml/operations-maintenance-database_maintenance.xml
```diff
@@ -0,0 +1,29 @@
...
+ </para>
+<screen>
+&prompt.ardana;cat /etc/cron.daily/local-cleanup-heat
+#!/bin/bash
+su heat -s /bin/bash -c "/usr/bin/heat-manage purge_deleted -g days 14" || :
+
```
jeremy-moffitt, line 18: its not clear what these commands do... or what the user would want to run them... this applies to all 3 commands here. We may want to spell out what the impact of running these commands is, and why someone would want to run them.

jeremy-moffitt (changes requested):
I think we need a little more information about what each of the commands does and why someone would choose to use them

xml/operations-maintenance-database_maintenance.xml
yarunachalam: I have removed

xml/operations-maintenance-database_maintenance.xml
yarunachalam: I will work on this to add details to each commands

rtamalin: So these are not commands
These are cron scripts that are run daily (at approx 04:02 effectively for most standard cron setups) on whichever node they are configured.

So you would need to document the process by which you configure these cron files on a system, and the process for determining which system you configure these commands on, and what dependencies there may be... E.g. the /usr/bin/heat-manage command can only be run on a system that has the appropriate conf settings to permit access to the DB. Similarly you likely need to be running the heat-manage command out of the related /opt/stack/services/<service_name>/bin on Ardana systems, unless we are creating a symlink from the venv files to /usr/bin/...

So are these cron scripts what you are proposing to setup to assist the on-going maintenance of the cloud? Or are they being installed by something on our systems already? If the latter then what is setting them up?

Also adding these scripts to the cron.daily means that every day the system will attempt to purge records that are older than X days... Which means that the first time it runs it will be purging lots of days with or records, but on subsequent days it will be purging just a single day's set of records... It may be better to add this to the cron.weekly or even cron.monthly so that it doesn't cause a regular daily "blip" in terms of DB access. The issue with purging such records is that "locking" the DB tables to do the purge is an expensive operation, using blocking other applications from accessing the DB, whereas once locked the time to purge 1 vs hundreds vs 1000s of records doesn't scale all that significantly. Millions of records may take some time... So this is more of a monthly maintenance operation, from my experience.

rtamalin (approved):
I think you need to determine if you are documenting the commands that can be run to perform the maintenance or the system configuration that can be performed to setup automated running of said maintenance tasks....

jeremy-moffitt:
superseded by https://github.com/SUSE-Cloud/doc-cloud/pull/1307

PR #1306, opened 2021-01-22
Add steps for archiving deleted instances from nova db (SOC-11403)

djoreilly:


jeremy-moffitt:
soc-11403
looks ok to me

jeremy-moffitt (approved):

PR #1307, opened 2021-02-22
Add nova and heat db purge cron jobs to maintenance section (SOC-9876)

jeremy-moffitt:
Adds nova and heat db purges on a weekly interval as recommended maintenance
soc-9876-1

### DIFF
--- a/xml/depl_maintenance.xml
+++ b/xml/depl_maintenance.xml
@@ -17,7 +17,7 @@
 </dm:docmanager>
 </info>
 <para/>
- <sect1 xml:id="sec-depl-maintenance-updates">
+ <section xml:id="sec-depl-maintenance-updates">
   <title>Keeping the Nodes Up-To-Date</title>
   <para>
    Keeping the nodes in &productname; up-to-date requires an appropriate
@@ -129,9 +129,9 @@
     </listitem>
    </varlistentry>
   </variablelist>
- </sect1>
+ </section>
 
- <sect1 xml:id="sec-depl-maintenance-service-order">
+ <section xml:id="sec-depl-maintenance-service-order">
   <title>Service Order on &cloud; Start-up or Shutdown</title>
 
   <para>
@@ -285,9 +285,9 @@
     </para>
    </listitem>
   </orderedlist>
- </sect1>
+ </section>
 
- <sect1 xml:id="sec-depl-maintenance-upgrade">
+ <section xml:id="sec-depl-maintenance-upgrade">
   <title>Upgrading from &productname; 7 to &productname; 8</title>
 
   <para>
@@ -317,7 +317,7 @@
      </para>
     </important>
 
-  <sect2 xml:id="sec-depl-maintenance-upgrade-require">
+  <section xml:id="sec-depl-maintenance-upgrade-require">
    <title>Requirements</title>
    <para>
     When starting the upgrade process, several checks are performed to
@@ -520,16 +520,16 @@
      </para>
     </listitem>
    </itemizedlist>
-  </sect2>
+  </section>
 
-  <sect2 xml:id="sec-depl-maintenance-postgre-mariadb-upgrade">
+  <section xml:id="sec-depl-maintenance-postgre-mariadb-upgrade">
    <title>Preparing &postgres;-Based &productname; 7 for Upgrade</title>
    <para>
     Upgrading &productname; 7 is only possible when it uses &mariadb; deployed
     with the Database &barcl;. This means that before you can proceed with
     upgrading &productname; 7, you must migrate &postgres; to &mariadb; first. The following description covers several possible scenarios.
    </para>
-   <sect3 xml:id="sec-postgre-mariadb-upgrade-scenario1">
+   <section xml:id="sec-postgre-mariadb-upgrade-scenario1">
     <title>Non-HA Setup or HA Setup with More Than 2 Nodes in the Cluster and
     &postgres; Database Backend</title>
     <para>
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
Note that this would supersede https://github.com/SUSE-Cloud/doc-cloud/pull/1301  ... I can add the fernet instruction from there, though I'm not clear on the specific motivation or what the IP addresses in that commit represent, and would need that information to add it.

rtamalin:
So is it intended to be weekly or daily? Your Description starts with "Adds nova and heat db purges on a weekly interval as recommended maintenance" but the scripts are being added to the cron.daily to run every day ;-)

rtamalin:
> So is it intended to be weekly or daily? Your Description starts with "Adds nova and heat db purges on a weekly interval as recommended maintenance" but the scripts are being added to the cron.daily to run every day ;-)

Ah I see you have already fixed this in code, but not in the rendered example ;-)

xml/depl_maintenance.xml
```diff
@@ -2181,4 +2181,30 @@ migrate</screen>
...
+  <para>
+    Heat-manage helps manage Heat specific database operations. The associated
+    database should be periodically purged to save space. The following should
+    be setup as a cron job at
+    <literal>/etc/cron.weekly/local-cleanup-heat</literal>
+    with the following content:
+  </para>
+  <screen>
+     #!/bin/bash
+     su heat -s /bin/bash -c "/usr/bin/heat-manage purge_deleted -g days 14" || :
```
rtamalin: Instead of `|| :` might it be better to use the less cryptic `|| true`?

Not sure where the reference example job definition came from, and it is perfectly valid. Just a thought...

xml/depl_maintenance.xml
```diff
@@ -2181,4 +2181,30 @@ migrate</screen>
...
+     nova-manage db archive_deleted_rows command will move deleted rows
+     from production tables to shadow tables. Including
+     <literal>--until-complete</literal> will make the command run continuously
+     until all deleted rows are archived. It is recommended to setup this task
+     as <literal>/etc/cron.weekly/local-cleanup-nova</literal> with the
+     following content:
+  </para>
+  <screen>
+      #!/bin/bash
+      su nova -s /bin/bash -c "/usr/bin/nova-manage db archive_deleted_rows --until-complete" || :
```
rtamalin: As above with regard to the `|| :`...

xml/depl_maintenance.xml
jeremy-moffitt: these came from the scripts that Bernhard runs (presumably on the ECP) and are quoted directly in the ticket, so I didnt want to mess with them

jeremy-moffitt:
soc-9876-2

jeremy-moffitt:
soc-9876-2

jeremy-moffitt:
soc-9876-3

xml/operations-maintenance-database_maintenance.xml
```diff
@@ -0,0 +1,48 @@
...
+     It is recommended to setup this task as a cron job in
+     <literal>/etc/cron.weekly/openstack-keystone-fernet</literal>
+     on the server where the keystone service is running,
+     with the following content:
+  </para>
+  <screen>
+  #!/bin/bash
+  su keystone -s /bin/bash -c "keystone-manage fernet_rotate"
+
+  /usr/bin/keystone-fernet-keys-push.sh 192.168.81.168; /usr/bin/keystone-fernet-keys-push.sh 192.168.81.169;
```
guangyee: this part is for Crowbar right? For Ardana, the keystone playbooks take care of that I think.

jeremy-moffitt: that makes sense... I'll see if I can set that to only render in the crowbar version

jeremy-moffitt:
soc-9876-4
removed fernet token instructions from CLM version

jeremy-moffitt:
For CLM, https://documentation.suse.com/soc/8/single-html/suse-openstack-cloud-operations/#fernet-tokens  already exists for the Fernet tokens

xml/depl_maintenance.xml
```diff
@@ -2180,5 +2180,22 @@ migrate</screen>
...
- </sect1>
+ </section>
+ <xi:include href="operations-maintenance-database_maintenance.xml"/>
+ <section xml:id="sec-depl-maintenance-fernet-tokens">
+  <title>Rotating Fernet Tokens</title>
+  <para>
+     Fernet tokens should be rotated frequently for security purposes.
+     It is recommended to setup this task as a cron job in
+     <literal>/etc/cron.weekly/openstack-keystone-fernet</literal>
+     on the server where the keystone service is running,
```
guangyee: Instead of "on the server where the keystone service is running", how about this?

"on the keystone server designated as a master node in a highly available setup, with the following content:"

xml/depl_maintenance.xml
```diff
@@ -2180,5 +2180,22 @@ migrate</screen>
...
+     with the following content:
+  </para>
+  <screen>
+  #!/bin/bash
+  su keystone -s /bin/bash -c "keystone-manage fernet_rotate"
+
+  /usr/bin/keystone-fernet-keys-push.sh 192.168.81.168; /usr/bin/keystone-fernet-keys-push.sh 192.168.81.169;
+  </screen>
```
guangyee, line 2199: Also add the following ...

"The IP addresses in the above example, i.e. 192.168.81.168 and 192.168.81.169, are the IP addresses of the other two nodes of a three-node cluster. If the other nodes have different IP addresses, you should use them accordingly."

jeremy-moffitt:
soc-9876-5

guangyee (approved):
Looks good to me!

PR #1310, opened 2021-04-27
Add instructions for checking MySQL cert expiry (SOC-11422)

jeremy-moffitt:
Add instructions for determining if the MySQL/Percona cluster has expired certs
see: https://bugzilla.suse.com/show_bug.cgi?id=1178831

bz1178831_cloud9

### DIFF
--- a/xml/security-troubleshooting_tls.xml
+++ b/xml/security-troubleshooting_tls.xml
@@ -104,6 +104,13 @@ ansible-playbook -i hosts/verb_hosts monasca-start.yml</screen>
   <procedure>
    <step>
     <para>
+      Determine if the TLS certificates for MySQL / Percsona have expired.
+    </para>
+    <screen>&prompt.ardana;cd /etc/mysql/
+&prompt.ardana;openssl x509 -noout -enddate -in control-plane-1-mysql-internal-cert.pem</screen>
+   </step>
+   <step>
+    <para>
      Regenerate the TLS certificates on the deployer.
     </para>
     <screen>&prompt.ardana;cd ~/scratch/ansible/next/hos/ansible


xml/security-troubleshooting_tls.xml
```diff
@@ -102,6 +102,14 @@ ansible-playbook -i hosts/verb_hosts monasca-start.yml</screen>
    Percona clusters.
   </para>
   <procedure>
+   <step>
+    <para>
+      Determine if the TLS certificates for MySQL / Percsona have expired.
+    </para>
+    <screen>&prompt.ardana;cd /etc/mysql/
+&prompt.ardana;/etc/mysql # openssl x509 -noout -text -in control-plane-1-mysql-internal-cert.pem | grep After
```
guangyee: To be precise, you may want to do `openssl x509 -noout -enddate -in control-plane-1-mysql-internal-cert.pem` instead.

jeremy-moffitt: updated, please take another look, thanks! (did not update the attached image, just the source file)

xml/security-troubleshooting_tls.xml
```diff
@@ -102,6 +102,14 @@ ansible-playbook -i hosts/verb_hosts monasca-start.yml</screen>
    Percona clusters.
   </para>
   <procedure>
+   <step>
+    <para>
+      Determine if the TLS certificates for MySQL / Percsona have expired.
+    </para>
+    <screen>&prompt.ardana;cd /etc/mysql/
+&prompt.ardana;/etc/mysql # openssl x509 -noout -enddate -in control-plane-1-mysql-internal-cert.pem | grep After
```
guangyee: you don't need the `| grep After` part as the `-enddate` option only prints out the end date attribute

guangyee: the output looks like this `notAfter=Jul 24 12:24:17 2021 GMT`

jeremy-moffitt:
trying to test this out and not finding the pem file where specified, probably user error on my part, but adding `do not merge yet` label until resolved

jeremy-moffitt:
bz1178831_cloud9
confirmed on a SOC9 install, removed the directory path because it looked awkward in the instructions
```
ardana-jmoffitt-tls-check-cp1-c1-m1-mgmt:/etc/mysql # openssl x509 -noout -enddate -in control-plane-1-mysql-internal-cert.pem
notAfter=May  4 21:08:42 2022 GMT
```

guangyee (approved):
looks good!

PR #1316, opened 2021-05-11
SOC-11522 -Add bm-power-status playbook in add SLES compute section

yarunachalam:
Add bm-power-status playbook in add SLES compute section (SOC-11522)


jeremy-moffitt:
think this change is meant to be targeted at `maintenance/cloud_8` rather than `maintenance/cloud_9` , closing PR

PR #1322, opened 2021-05-11
Update the correct SLES version to suse-12.3 (SOC-11521)

yarunachalam:
Update the correct SLES version to suse-12.3 (SOC-11521)

### DIFF
--- a/xml/depl_inst_nodes.xml
+++ b/xml/depl_inst_nodes.xml
@@ -403,7 +403,7 @@
     <para>
      Download the <filename>crowbar_register</filename> script from the
      &admserv; at
-     <literal>http://<replaceable>192.168.124.10</replaceable>:8091/suse-12.2/x86_64/crowbar_register</literal>.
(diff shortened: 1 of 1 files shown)


xml/depl_inst_nodes.xml
```diff
@@ -403,7 +403,7 @@
     <para>
      Download the <filename>crowbar_register</filename> script from the
      &admserv; at
-     <literal>http://<replaceable>192.168.124.10</replaceable>:8091/suse-12.2/x86_64/crowbar_register</literal>.
```
jeremy-moffitt, line 406: I'm not sure about this change, SOC9 is SLES12 SP4... though the docs previously said SP3? strange... have you validated the url?

yarunachalam: it should be SP4

yarunachalam: Resolved

jeremy-moffitt (approved):

PR #1326, opened 2021-07-20
SOC 9 DC files: Update to 2021 stylesheets

ghost:
This updates the SOC 9 docs to the 2021 SUSE branding.

Be aware that if you intend to ship another documentation package update as a MU, this has the potential to break your build. However, @tomschr has already updated the suse-xsl-stylesheets package for most SLE distributions, so I think you should be fine even if you need to make a doc MU.

There's a related PR for SOC 8, see #1327

PR #1327, opened 2021-07-20
SOC 8 DC files: Update to 2021 stylesheets

ghost:
This updates the SOC 8 docs to the 2021 SUSE branding.

Be aware that if you intend to ship another documentation package update as a MU, this has the potential to break your build. However, @tomschr has already updated the suse-xsl-stylesheets package for most SLE distributions, so I think you should be fine even if you need to make a doc MU.

There's a related PR for SOC 8, see #1326


jeremy-moffitt (approved):
built locally, looks good, approved and merging

ghost:
Thank you!

PR #1328, opened 2021-08-04
Add SES7 integration instructions (SOC-11397)

jeremy-moffitt:
Add SES7 integration instructions for both Crowbar and CLM


jeremy-moffitt:
trying to figure out why the validation is failing, there's probably a way to run the geeko validation locally but I dont have that handy and it passes daps2docker regular validation, so there may be a fair number of force pushes as I debug

jeremy-moffitt:
the issue was having a `screen` inside of an `important` ... fixed, not sure I love the result, but its o.k..

djz88 (approved):
LGTM

PR #1331, opened 2021-11-12
Fix dm:editurls to improve broken-link situation

ghost:


jeremy-moffitt (approved):
build both CLM and Crowbar locally with these changes as a sanity check and both looked ok to me.

ghost:
Thanks for the review! With that, I merged!

PR #1332, opened 2022-01-04
Add instructions for updating keystone has algorithm (bsc#1186380)

jeremy-moffitt:
Adds steps for users who want to change the configured hashing algorithm for keystone

### DIFF
--- a/xml/operations-configuring-identity-identity_limitations.xml
+++ b/xml/operations-configuring-identity-identity_limitations.xml
@@ -90,6 +90,51 @@
     </para>
    </listitem>
   </itemizedlist>
+  <para>
+   <emphasis role="bold">Hash Algorithm Configuration</emphasis>
+  </para>
+  <itemizedlist>
+    <listitem>
+     <para>
+     The default hash algorithm is <literal>bcrypt</literal>, which has a built-in
+     limitation of 72 characters. As keystone defaults to a secret length of 86
+     characters, customers may choose to change the keystone hash algorithm to one
+     that supports the full length of their secret.
+    </para>
+   </listitem>
+   <listitem>
+     <para>Process for changing the hash algorithm configuration:</para>
+     <procedure>
+       <step>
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
bz1186380

xml/operations-configuring-identity-identity_limitations.xml
```diff
@@ -90,6 +90,51 @@
...
+   <listitem>
+     <para>Process for changing the hash algorithm configuration:</para>
+     <procedure>
+       <step>
+         <para>
+           Update the <literal>identity</literal> section of keystone.conf.j2 to reference the desired algorithm
+         </para>
+<screen>
+[identity]
+password_hash_algorithm=sha512_crypt
```
guangyee: We should use "pbkdf2_sha512" instead of "sha512_crypt" here as "sha512_crypt" is not in the supported config.
See https://opendev.org/openstack/keystone/src/branch/stable/rocky/keystone/conf/identity.py#L113

jeremy-moffitt: good catch, I will fix that, rebuild and repush... I may also squash the cherry-pick into the base commit

jeremy-moffitt:
bz1186380
updated picture

xml/operations-configuring-identity-identity_limitations.xml
jeremy-moffitt: fix pushed

guangyee (approved):
looks good!

KeithMnemonic (approved):
LGTM, but is the update in the soc-operations-self-sign-cert.xml  related to the HASH?

jeremy-moffitt:
> LGTM, but is the update in the soc-operations-self-sign-cert.xml related to the HASH?

It is not, there was a change to the validation schema for docbook xml and I tweaked that file to pass the test.

PR #1333, opened 2022-04-04
Add DOCBOOK5_RNG_URI in DC files to validate

tomschr:
Some guides did not validate with GeekoDoc. As SOC 8 is now unsupported, it doesn't make sense to go through all the files and change them.

To be able to build and publish them on docserv on last time, we need to ensure we validate with DocBook 5.1. This PR make this happen.


fsundermeyer (approved):
LGTM

jeremy-moffitt (approved):
I built it locally and it looks good. Thanks for being proactive about this!

PR #1334, opened 2022-04-06
Use old 2013-ns stylesheet to build old DB4 sources

tomschr:
When I tried to build SOC8, it failed. It was due to that our newer stylesheets requires DocBook 5, but some SOC8 sources uses still DocBook 4.

This PR changes the new stylesheets to the suse2013-ns stylesheets which supports DocBook 4. No content changes.


jeremy-moffitt (approved):
built clean, spot check looked good... :+1:

PR #1341, opened 2022-10-17
Change suse stylesheets for SOC 8

tomschr:
# Problem

The current SOC 8 HTML have some issues regarding SEO. The old stylesheets (suse2021-ns) doesn't support them and won't get an update.

# Solution
This PR doesn't have any content changes. It only changes the DC files and switch to our new layout (suse2022-ns).


jeremy-moffitt (approved):
this is fine, but I dont think we're publishing SOC8 docs publicly anymore... not sure if its worth updating them...

